PR #2, opened 2017-07-26
Automate tests

MrCreosote:
Adds test code for automating the tests. Set up so
the tests can be run as part of kb-sdk test or standalone (e.g. in
eclipse). Defaults for test properties are set so they work with kb-sdk
test.

This PR sits on top of #1.

TODO:
* [x] Don't use a prefab workspace to run tests, create on startup
* [x] Run workspace locally rather than using CI
* [ ] Determine a fix for the Too many dynamic script compilations within one minute error - *not getting done for this PR*
* [x] Automate the rest of the mongodb related tests
* [x] Automate the ES related tests
* [x] Add an ElasticSearch controller
* [x] Figure out why the java process hangs after a successful test run.


MrCreosote:
java hang was fixed by either a) updating VirtualBox to 5.1.24, or b) switching Docker from AUFS to OverlayFS.

MrCreosote:
@bill this is as ready to merge as it's going to get. There's still one test that fails because it exceeds ElasticSearch limits, but I haven't been able to get resolution on whether that's a problem with the code or that I need to boost the limits in the ES config. Since I've been waiting for several days we should go ahead and merge.

Dockerfile
```diff
@@ -22,6 +22,23 @@ RUN sudo apt-get install nano \
...
+	&& tar xfz elasticsearch-5.5.0.tar.gz \
+	&& ln -s elasticsearch-5.5.0 elasticsearch
+	
+	
+RUN cd /opt \
+	&& wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-2.6.12.tgz \
+	&& tar xfz mongodb-linux-x86_64-2.6.12.tgz \
+	&& ln -s mongodb-linux-x86_64-2.6.12 mongo
+
+# change mrcreosote -> kbase when PR is merged
```
briehl: Do this change before or after the merge?

test/src/kbaserelationengine/main/test/MainObjectProcessorTest.java
```diff
@@ -23,56 +32,90 @@
...
+        mc = new MongoClient("localhost:" + mongo.getServerPort());
+        final String dbName = "DataStatus";
+        db = mc.getDatabase(dbName);
+        
+        // set up elastic search
+        es = new ElasticSearchController(TestCommon.getElasticSearchExe(), tempDir);
+        
+        // set up Workspace
+        ws = new WorkspaceController(
+                "0.7.2-dev1",
```
briehl: Maybe make this configurable so you don't have to dig later?

Dockerfile
MrCreosote: After. This'll work fine for now, the change is really just for looks. It's just waiting on this PR: https://github.com/kbase/jars/pull/77

briehl: :+1:

test/src/kbaserelationengine/main/test/MainObjectProcessorTest.java
MrCreosote: Sure, I can do that, although this is currently the only version of the workspace that'll work with RESKE once it's updated to use the admin commands. I'll do it in the next PR if that's ok with you.

MrCreosote:
Bill said ok to merge as is.

PR #5, opened 2017-08-17
Fix script complication fails

MrCreosote:
The script complication circuit breaker activates *a lot* when cloning a
large workspace (cloning code is in another branch ATM). This PR fixes the
scripts to only need to compile once when first used, rather than
possibly, depending on the parameters, on every invocation.

There are now up to 10 script compilations possible as long as the scripts stay cached,
so the compliations / min circuit breaker should no longer be tripped.

Tests now pass.

TODO:
* [x] fix or rework the final script


MrCreosote:
Same build fail as https://github.com/kbaseapps/KBaseRelationEngine/pull/4#issuecomment-322296597

PR #7, opened 2017-08-23
TASK-895 Add access group id cache & workspace provider implementation

MrCreosote:
Adds a cache for access group IDs per user. Expires based on
configurable time in cache and cache size. Cache is source agnostic and
wraps and presents the same AccessGroupProvider interface.

Now queries the workspace for private access groups rather than maintaining a local copy.

Tested manually.

On extremely limited data, search appears to be pretty fast - 31ms on a cache miss vs. 15ms on a cache hit when running RESKE and the workspace locally.

TODO:
* [x] use the cache
* [x] implement a workspace access provider and use


MrCreosote:
@rsutormin this is ready for merge

PR #9, opened 2017-08-24
TASK-895 Add delete object event handling.

MrCreosote:


lib/src/kbaserelationengine/main/MainObjectProcessor.java
```diff
@@ -444,11 +450,20 @@ public void indexObject(GUID guid, String storageObjectType, Long timestamp, boo
     public void share(GUID guid, int accessGroupId) throws IOException {
         indexingStorage.shareObjects(new LinkedHashSet<>(Arrays.asList(guid)), accessGroupId);
     }
+    
+    public void shareAllVersions(final GUID guid) throws IOException {
+//        indexingStorage.shareAllVersions(guid);
```
rsutormin: Here is what's needed to share all versions:
method `loadLastVersion` (like here https://github.com/kbaseapps/KBaseRelationEngine/blob/master/lib/src/kbaserelationengine/search/ElasticIndexingStorage.java#L236) but with `null` in last parameter
and then `updateLastVersionsInData` (like here https://github.com/kbaseapps/KBaseRelationEngine/blob/master/lib/src/kbaserelationengine/search/ElasticIndexingStorage.java#L252).

PR #11, opened 2017-08-25
Support for DataPalettes in public workspaces

rsutormin:


rsutormin:
@MrCreosote: I think this PR is ready for review.

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
```diff
@@ -1064,18 +1192,36 @@ public FoundHits searchObjects(String objectType, MatchFilter matchFilter,
...
+        should.add(new LinkedHashMap<String, Object>() {{
             put("terms", match);
+        }});
+        if (withPublic) {
+            // Case of public workspaces containing DataPalette referencing to given object 
+            Map<String, Object> exists = new LinkedHashMap<String, Object>() {{
+                put("field", "extpub");
+            }};
+            should.add(new LinkedHashMap<String, Object>() {{
+                put("exists", exists);
```
MrCreosote: Need an explanation of what's going on here - looks like it's just checking for the existence of the pub ids list? Isn't that always going to exist?

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
```diff
@@ -788,14 +811,115 @@ public void deleteAllVersions(final GUID guid) throws IOException {
...
+        }};
+        String urlPath = "/" + indexName + "/" + getAccessTableName() + "/_update_by_query";
+        Response resp = makeRequest("POST", urlPath, doc);
+        Map<String, Object> data = UObject.getMapper().readValue(
+                resp.getEntity().getContent(), Map.class);
+        return (Integer)data.get("updated") > 0;
+    }
+
+    @Override
+    public void publishObjectsExternally(Set<GUID> guids, int accessGroupId) throws IOException {
```
MrCreosote: This and unpublish aren't actually used, are they? They still need to be wired up to the event handler?

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
```diff
@@ -1306,6 +1452,10 @@ private void createAccessTable(String indexName, Map<String, Object> mappings) t
         props.put("groups", new LinkedHashMap<String, Object>() {{
             put("type", "integer");
         }});
+        // List of external workspaces containing DataPalette pointing to this object
+        props.put("extpub", new LinkedHashMap<String, Object>() {{
+            put("type", "integer");
+        }});
```
MrCreosote: is the -1 flag still relevant?

MrCreosote:
Also, does deleteAllVersions need to do anything different now?

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
rsutormin: They should but it's more tricky and I suggest to do it in separate PR. What do you think?

MrCreosote: Sure

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
rsutormin: No, that's an essential parameter that you want to check: how many public workspaces (external comparing to home workspace of given object) have DataPalettes referencing given object (and version). If this number is 0 then object+version is not visible as public through DataPalettes. If it's >0 then it's visible.

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
rsutormin: I'll remove -1 in separate PR. It's not used and doesn't do any harm anyway.

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
MrCreosote: I see, the exists query isn't just that the field exists, it's that it exists and contains at least one non-null entry, so an empty list will be false

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
MrCreosote: ok

PR #13, opened 2017-08-25
TASK-894 Use the event storage API rather than contacting mongo directly + ws blacklist

MrCreosote:
Instead of placing records directly in mongo, use the event storage API.

Would prefer to do something similar with workspace but dependency hell prevents that for now.

Also updated the WS storage code to use the 3.0 mongo API.

Also also added a workspace blacklist config & CLI parameter.

### DIFF
--- a/lib/src/kbaserelationengine/tools/RESKETools.java
+++ b/lib/src/kbaserelationengine/tools/RESKETools.java
@@ -9,6 +9,7 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import java.util.Properties;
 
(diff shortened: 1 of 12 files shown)


lib/src/kbaserelationengine/tools/RESKETools.java
```diff
@@ -99,17 +100,20 @@ public int execute() {
             printError("For config file " + a.configPath, e, a.verbose);
             return 1;
         }
-        return runEventGenerator(cfg, out, a.ref, a.verbose);
+        final List<Integer> wsBlackList = a.wsBlacklist == null || a.wsBlacklist.isEmpty() ?
```
rsutormin: For higher convenience we can add support for workspace names in black list as well (maybe later).

PR #15, opened 2017-08-29
TASK-1014 WIP Add object fields to index

MrCreosote:
Add selected fields from object_info and provenance

* Added some refactoring to remove workspace specific code from the main loop
  * Ideally eventually the workspace related code will be completely hidden inside the workspace event handler
* Remove unused metadata field for now.
* Added SourceData class to package information about the data and pass it around. Finally winds up in the indexer and gets indexed.
* Add properties from object_info and object provenance to the SourceData and index them.


MrCreosote:
@rsutormin  this is ready for review.

lib/src/kbaserelationengine/search/ElasticIndexingStorage.java
```diff
@@ -211,7 +219,15 @@ public void indexObjects(String objectType, String objectName, long timestamp,
         doc.putAll(indexPart);
         doc.put("guid", id.toString());
         doc.put("otype", objectType);
-        doc.put("oname", objectName);
+
+        doc.put("oname", data.getName());
+        doc.put("creator", data.getCreator());
```
rsutormin: I think you should add description for new fields to Elastic data structure definition code.

MrCreosote: Good catch, fixed. Ready for review again.

PR #17, opened 2017-08-31
TASK-895 Add undelete object event & fix delete object

MrCreosote:
Not clear how to handle admin access group on a delete


MrCreosote:
@rsutormin added

PR #30, opened 2017-09-19
Refactor MOP to use dependency injection, fix workspace timestamp

MrCreosote:
* Use the object timestamp instead of the event timestamp when indexing objects
* Switch to dependency injection for MOP creation. Should make splitting the indexer out of MOP much easier. Also makes testing and component swapping easier.


MrCreosote:
Merging per @scanon @sychan

PR #47, opened 2017-10-16
Refactor statusevent w/ builder

MrCreosote:
Adds a builder for status event in preparation for removing the id field
and adding a processing status field.

Also switch long, which is ambiguous, to Instant, and use Optionals for
nullable values.


MrCreosote:
Merging per @sychan

PR #48, opened 2017-10-16
TASK-1161 Store sub events in database

MrCreosote:
Now stores sub events in database and generates an ID for the sub event.
Does not yet check to avoid storing duplicate events or reprocessing
successful sub events.

Removed the id field from StatusEvent (now in StatusEventWithID).

Remove a not particularly helpful level of indirection by marking event
processing state in the storage system with a direct storage call,
rather than via an iterator then a queue then the storage system


MrCreosote:
Merging per @sychan

PR #49, opened 2017-10-17
TASK-1161 Expand event processing state model to mark fails

MrCreosote:
Previous model w/ 2 bools was unprocessed, unindexed, and indexed

New model with one expandable enum adds fail


MrCreosote:
Requires https://github.com/kbase/workspace_deluxe/pull/177

PR #50, opened 2017-10-18
Add tests to status event classes

MrCreosote:


arfathpasha:
For consistency, it looks like "isGlobalAccessed" should be changed to "isPublic" on the lines below?

https://github.com/kbase/KBaseSearchEngine/blob/master/lib/src/kbasesearchengine/events/storage/OldMongoDBStatusEventStorage.java#L57
https://github.com/kbase/KBaseSearchEngine/blob/master/lib/src/kbasesearchengine/events/storage/OldMongoDBStatusEventStorage.java#L98

MrCreosote:
@arfathpasha That whole class is going to be rewritten shortly, so for the purposes of this pr it's moot.

PR #52, opened 2017-10-19
TASK-1161 Move processing state out of StatusEvent, tests

MrCreosote:
StatusEvent shouldn't contain a processing state that mutates over time (even though the class itself is immutable)


arfathpasha:
- If there is no status/processingState in StatusEvent, should it just be called Event? Event gels well with getEvent() in StoredStatusEvent. Description would then be "An event to which the indexer should respond."
- StoredStatusEvent can then be called StoredEventStatus or perhaps just EventStatus
- StatusEventProcessingState can be simplified to EventState

MrCreosote:
The status in statusevent refers to status changes in an external data storage system, not the processing status of the event itself.

PR #55, opened 2017-10-24
TASK-1161 Rewrite storage system

MrCreosote:
This PR is way too big, but we're swapping out the storage system and
adding tests, so...

Besides a rewrite of the storage system & tests also fixed a startup bug
when starting the indexer coordinator


MrCreosote:
@arfathpasha Here's the workspace side changes: https://github.com/kbase/workspace_deluxe/pull/180

PR #63, opened 2017-11-03
TASK-1206 Add ObjectParse unit test

Tianhao-Gu:
Add tests for OjbectParse class


test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
+package kbasesearchengine.test.parse;
+
```
MrCreosote: What coverage does this test have for the class and its dependencies? For immediate dependencies I see:
SubObjectExtractor
SimpleSubObjectConsumer
IdMapper
SimpleSubObjectConsumer
ObjectJsonPath

Those may have further dependencies.
In some cases it may make more sense to refactor the class so the dependencies can be injected rather than embedded (especially for static method calls).

MrCreosote: Can the API be reduced - e.g. are there methods that can be made private to reduce the size of the API? Are there any unused variables in the method signatures that could be removed?

MrCreosote: Speaking of method sigs, this is a good time to clean them up. My rules are:
1) all inputs are `final` unless there's a really good reason not to be
2) if the signature can't entirely fit on one line then:
  a) Move the throws statement to the next line, indent an extra 8 spaces. If another line for exceptions is needed, indent an additional 4.
  b) If the sig still can't fit on one line then put each variable on its own line.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+		impl = new ObjectParser();
+	}
+
+	@AfterClass
+	public static void teardown() throws Exception {
+		FileUtils.deleteDirectory(scratch.toFile());
+	}
+
+	@Test
+	public void prepareTempFileTest() throws Exception {
```
MrCreosote: I'm not sure why the temp file prep method belongs in ObjectParser - it has nothing to do with the rest of the class. Seems like it should be moved into the IndexerWorker class, maybe.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
+package kbasesearchengine.test.parse;
```
MrCreosote: As part of writing the tests, I was thinking that we should be refactoring the code to be understandable, documented, and following good practices such as not returning nulls (and avoiding nulls in general) and that kind of thing.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	@SuppressWarnings("static-access")
+	@Test
+	public void extractParentFragmentGenomeFeatureTest() throws Exception {
+
+		String jsonResource = "genome01";
+		String type = "GenomeFeature";
+
+		String parentJson = this.extractParentFragment(type, jsonResource);
```
MrCreosote: You don't need `this` to call a static method do you?

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+		String jsonResource = "genome01";
+		String type = "GenomeFeature";
+
+		List<String> expectedIds = Arrays.asList("NewGenome.CDS.6211", "NewGenome.CDS.6210", "NewGenome.CDS.6209");
+
+		ArrayList<GUID> idList = this.prepareGUID(type, jsonResource, new GUID("WS:1/1/1"));
+
+		assertThat(idList.size(), is(3));
+		for (GUID guid : idList) {
+			assertThat(expectedIds, hasItem(guid.getSubObjectId()));
```
MrCreosote: It there a reason you're not checking the ordering of the results?

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+	public static Map<GUID, String> parseSubObjects(String type, String jsonResource, String guidString)
+			throws Exception {
+
+		GUID guid = new GUID(guidString);
+
+		InputStream inputStream = ObjectParserTest.class.getResourceAsStream(jsonResource + ".json.properties");
+		Reader reader = new InputStreamReader(inputStream);
+
+		UObject data = UObject.fromJsonString(CharStreams.toString(reader));
+        String name = "TestObj" + "_" + type;
```
MrCreosote: Looks like this is mixing tabs and spaces? For this repo we're going with spaces, 4 per indent.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+        String creator = "creator";
+        
+        SourceData obj = SourceData.getBuilder(data, name, creator).withNullableCommitHash(null)
```
MrCreosote: No need to call `.with` methods with a null value, it's a noop

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+		String type = "GenomeFeature";
+		String guid = "WS:1/1/1";
+
+		Map<GUID, String> guidToJson = this.parseSubObjects(type, jsonResource, guid);
+
+		assertThat(guidToJson.size(), is(3));
+
+		List<String> expectedIds = Arrays.asList("NewGenome.CDS.6211", "NewGenome.CDS.6210", "NewGenome.CDS.6209");
+		for (GUID guidKey : guidToJson.keySet()) {
+			assertThat(expectedIds, hasItem(guidKey.getSubObjectId()));
```
MrCreosote: Same question about ordering

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+		@SuppressWarnings("static-access")
+		File tmpFile = impl.prepareTempFile(tempDir);
+		assertEquals(true, tempDir.exists());
+		assertEquals(true, tmpFile.exists());
+		
+		assertThat(tmpFile.toString(), both(containsString("ws_srv_response_")).and(containsString(".json")));
+	}
+
+	@SuppressWarnings("static-access")
```
MrCreosote: What's causing this warning? I've never had to use it.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+	public void extractParentFragmentGenomeTest() throws Exception {
+
+		String jsonResource = "genome01";
+		String type = "Genome";
+
+		String parentJson = this.extractParentFragment(type, jsonResource);
+		assertNull(parentJson); // Genome.json doesn't have "from-parent" path
+	}
+
+	@SuppressWarnings("unchecked")
```
MrCreosote: Rather than marking an entire method as unchecked, just mark the variable(s).

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+			assertThat(guidKey.getSubObjectType(), is("feature"));
+			assertThat(guidKey.getStorageCode(), is("WS"));
+			assertThat(guidKey.getVersion(), is(1));
+			assertThat(guidKey.getAccessGroupId(), is(1));
+			assertThat(guidKey.getAccessGroupObjectId(), is("1"));
+		}
+
+		List<String> expectedKeys = Arrays.asList("ontology_terms", "function", "protein_translation", "location", "id", "type");
+		for (String jsonString : guidToJson.values()) {
+			HashMap<String, Object> result = new ObjectMapper().readValue(jsonString, HashMap.class);
```
MrCreosote: Why doesn't this check the key values?

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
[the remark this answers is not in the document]
Tianhao-Gu: Yea. It's added as '<include name="kbasesearchengine/test/parse/**Test.java"/>'

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
+package kbasesearchengine.test.parse;
+
+import static org.hamcrest.CoreMatchers.is;
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,256 @@
...
+public class ObjectParserTest {
+
+	private static Map<String, String> config = null;
+	private static Path scratch;
+	private static ObjectParser impl = null;
+
+	@BeforeClass
+	public static void init() throws Exception {
+		// config loading
+		String configFilePath = System.getenv("KB_DEPLOYMENT_CONFIG");
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
Tianhao-Gu: 👍

Tianhao-Gu:
@MrCreosote Could you take another look at this PR when you have a minute? Thanks.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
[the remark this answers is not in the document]
MrCreosote: gotcha

MrCreosote: I still think this one needs some work right?

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
MrCreosote: Same here

MrCreosote: same here

MrCreosote: I don't think all the sigs fit these rules yet?

PR #65, opened 2017-11-09
Add object level queue

MrCreosote:
An access group level queue will wrap, initialize, and control multiple object level queues.

This is really big, but it's almost all greenfield code and so is easier to read.

Either @arfathpasha or @tianhao-gu could review this, but tian doesn't have write access to I can't add him to the reviewer list.


sychan:
@Tianhao-Gu @MrCreosote I've added Tian to the ACL for this repo

PR #66, opened 2017-11-10
TASK-1160 Add ability for external actor to block queue

MrCreosote:


MrCreosote:
Hmm, it looks like the travis tests still aren't working, even after @sychan 's recent pr.

MrCreosote:
@arfathpasha did you set up the travis stuff? seems like the test.cfg file isn't getting properly initialized. 

One thing I notice is that we're catting test.cfg after the tokens are supposed to have been decrypted and inserted, which seems like a Bad Thing.

PR #69, opened 2017-11-16
TASK-1196 - Rules validation

arfathpasha:
Need to validate inputs before writing tests for keyword parser.

### DIFF
--- a/lib/src/kbasesearchengine/parse/KeywordParser.java
+++ b/lib/src/kbasesearchengine/parse/KeywordParser.java
@@ -19,6 +19,7 @@ import kbasesearchengine.events.exceptions.IndexingException;
 import kbasesearchengine.search.ObjectData;
 import kbasesearchengine.system.IndexingRules;
 import kbasesearchengine.system.ObjectTypeParsingRules;
+import kbasesearchengine.system.ValidationException;
 import us.kbase.common.service.UObject;
 
 public class KeywordParser {
@@ -33,6 +34,19 @@ public class KeywordParser {
             final ObjectLookupProvider lookup,
             final List<GUID> objectRefPath)
             throws IOException, ObjectParseException, IndexingException, InterruptedException {
+
+        // check pre-conditons
+        if(json == null) throw new IllegalArgumentException("json is a required parameter");
+        if(indexingRules == null) throw new IllegalArgumentException("indexingRules is a required parameter");
+
+        for(IndexingRules rule: indexingRules) {
+            try {
+                rule.validate();
+            } catch(ValidationException ex) {
+                throw new IllegalArgumentException("Unable to extract keywords", ex);
+            }
+        }
+
         Map<String, InnerKeyValue> keywords = new LinkedHashMap<>();
         ValueConsumer<List<IndexingRules>> consumer = new ValueConsumer<List<IndexingRules>>() {
             @Override
--- a/lib/src/kbasesearchengine/parse/KeywordParser.java
+++ b/lib/src/kbasesearchengine/parse/KeywordParser.java
@@ -20,6 +20,7 @@ import kbasesearchengine.search.ObjectData;
 import kbasesearchengine.system.IndexingRules;
 import kbasesearchengine.system.ObjectTypeParsingRules;
 import kbasesearchengine.system.ValidationException;
+import kbasesearchengine.tools.Utils;
 import us.kbase.common.service.UObject;
 
 public class KeywordParser {
@@ -36,13 +37,14 @@ public class KeywordParser {
             throws IOException, ObjectParseException, IndexingException, InterruptedException {
 
         // check pre-conditons
-        if(json == null) throw new IllegalArgumentException("json is a required parameter");
-        if(indexingRules == null) throw new IllegalArgumentException("indexingRules is a required parameter");
+        Utils.notNullOrEmpty(json, "json is a required parameter");
 
-        for(IndexingRules rule: indexingRules) {
+        Utils.nonNull(indexingRules, "indexingRules is a required parameter");
+
+        for (IndexingRules rule: indexingRules) {
             try {
                 rule.validate();
-            } catch(ValidationException ex) {
+            } catch (ValidationException ex) {
                 throw new IllegalArgumentException("Unable to extract keywords", ex);
             }
         }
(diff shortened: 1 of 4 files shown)


lib/src/kbasesearchengine/parse/KeywordParser.java
```diff
@@ -33,6 +34,19 @@ public static ParsedObject extractKeywords(
             final ObjectLookupProvider lookup,
             final List<GUID> objectRefPath)
             throws IOException, ObjectParseException, IndexingException, InterruptedException {
+
+        // check pre-conditons
+        if(json == null) throw new IllegalArgumentException("json is a required parameter");
```
MrCreosote: Looks like we have some style differences we should resolve for the sake of consistency in the codebase:
* I leave spaces after keywords so they don't look like fn calls (if here and for and catch below)
* I never use bracketless blocks, even when the entire if statement is on one line (goto fail bug-ish)

MrCreosote: You can use Utils.nonNull() for this. You might want to use Utils.notNullorEmpty() though.

lib/src/kbasesearchengine/parse/KeywordParser.java
```diff
@@ -33,6 +34,19 @@ public static ParsedObject extractKeywords(
...
+        if(indexingRules == null) throw new IllegalArgumentException("indexingRules is a required parameter");
+
+        for(IndexingRules rule: indexingRules) {
+            try {
+                rule.validate();
```
MrCreosote: I assuming we'll do the validation when we build the rules at some point rather than every time this method is called

lib/src/kbasesearchengine/parse/KeywordParser.java
MrCreosote: I'm curious to know if it's really necessary to pass around json strings rather than objects. Does the code keep a ton of strings in memory at once? I'm assuming the strings aren't that big...?

lib/src/kbasesearchengine/system/IndexingRules.java
```diff
@@ -2,23 +2,186 @@
...
+     */
     private String uiLinkKey = null;
+
+    /** Checks if this indexing rule is valid.
+     *
+     * @throws ValidationException if these rules are found to be invalid.
+     */
+    public void validate() throws ValidationException {
+        if(!derivedKey && path == null)
+            throw  new ValidationException("Must specify either derivedKey=true "+
```
MrCreosote: Another style question: spaces around operators

lib/src/kbasesearchengine/system/IndexingRules.java
```diff
@@ -2,23 +2,186 @@
...
+                    "sourceKey or path, but not both: "+toString());
+        if(derivedKey && sourceKey == null)
+            throw new ValidationException("derivedKey is true and source-key is " +
+                    "null expecting a non-null sourceKey to form the derived: " +
+                    "keyword from. "+toString());
+        if(fullText && keywordType != null)
+            throw new ValidationException("Specify either fullText=true or " +
+                    "sourceKey, but not both: "+toString());
+        if(transform != null && !( transform.startsWith("location") ||
+                transform.startsWith("length") || transform.startsWith("values") ||
```
MrCreosote: At some point these should be moved to constant fields

lib/src/kbasesearchengine/system/ValidationException.java
```diff
@@ -0,0 +1,24 @@
+package kbasesearchengine.system;
+
+/**
```
MrCreosote: add description of when this is thrown

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: Fixed

arfathpasha: I don't think there are that many json strings being held in memory and they can be represented as objects. I have made a note to look closer at this and do the refactoring for this a little down the line.

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: I have added a validation check at the building stage as well.

MrCreosote:
Need to fix the travis build at some point...

lib/src/kbasesearchengine/system/IndexingRules.java
```diff
@@ -2,23 +2,205 @@
...
+ *
+ * Each indexing rule forms a key-value pair based on some
+ * content of the object (or sub-object) that it is to be indexed by the rules.
+ * An indexing rule may also be based on a key-value pair formed by another
+ * indexing rule.
+ *
+ * See {@link #validate()} for instructions on building unambigious indexing rules.
+ *
+ */
 public class IndexingRules {
```
MrCreosote: I'm not saying do this now, but at some point we should move to a builder for an immutable class so the validate happens on a build.

lib/src/kbasesearchengine/parse/KeywordParser.java
```diff
@@ -33,6 +35,20 @@ public static ParsedObject extractKeywords(
...
             throws IOException, ObjectParseException, IndexingException, InterruptedException {
+
+        // check pre-conditons
+        Utils.notNullOrEmpty(json, "json is a required parameter");
+
+        Utils.nonNull(indexingRules, "indexingRules is a required parameter");
+
+        for (IndexingRules rule: indexingRules) {
+            try {
+                rule.validate();
```
MrCreosote: is this still needed given the validate in the objectparsingrules?

lib/src/kbasesearchengine/system/IndexingRules.java
```diff
@@ -2,23 +2,205 @@
...
+        if (derivedKey && sourceKey == null) {
+            throw new ValidationException("derivedKey is true and source-key is " +
+                    "null expecting a non-null sourceKey to form the derived: " +
+                    "keyword from. " + toString());
+        }
+        if (fullText && keywordType != null) {
+            throw new ValidationException("Specify either fullText=true or " +
+                    "sourceKey, but not both: " + toString());
+        }
+        if (transform != null && !( transform.startsWith("location") ||
```
MrCreosote: space inside parens

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: Until I can refactor to make an immutable class with builder and there is only one way to instantiate. Currently an instance can be created and morphed in many ways.

PR #70, opened 2017-11-16
TASK-1160 Add access group level queue

MrCreosote:
Test coverage is 100%

This is another big PR, but it's mostly greenfield again and so easier to read.

### DIFF
--- a/lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
+++ b/lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
@@ -79,11 +79,11 @@ public class AccessGroupEventQueue {
         final Map<String, List<StoredStatusEvent>> objectVersionProcessing = new HashMap<>();
         
         for (final StoredStatusEvent e: initialLoad) {
-            if (!e.getState().equals(StatusEventProcessingState.READY) ||
+            if (!e.getState().equals(StatusEventProcessingState.READY) &&
                     !e.getState().equals(StatusEventProcessingState.PROC)) {
                 throw new IllegalArgumentException("Illegal initial event state: " + e.getState());
             }
-            if (ACCESS_GROUP_EVENTS.contains(e)) {
+            if (ACCESS_GROUP_EVENTS.contains(e.getEvent().getEventType())) {
                 initAccessGroupEvent(e);
             } else if (ObjectEventQueue.isVersionLevelEvent(e)) {
                 initVersionEvent(objectVersionReady, objectVersionProcessing, e);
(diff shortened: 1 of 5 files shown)


lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
```diff
@@ -0,0 +1,314 @@
...
+        Utils.nonNull(initialLoad, "initialLoad");
+        Utils.noNulls(initialLoad, "initialLoad has null entries");
+        final Map<String, StoredStatusEvent> objects = new HashMap<>();
+        final Map<String, List<StoredStatusEvent>> objectVersionReady = new HashMap<>();
+        final Map<String, List<StoredStatusEvent>> objectVersionProcessing = new HashMap<>();
+        
+        for (final StoredStatusEvent e: initialLoad) {
+            if (!e.getState().equals(StatusEventProcessingState.READY) &&
+                    !e.getState().equals(StatusEventProcessingState.PROC)) {
+                throw new IllegalArgumentException("Illegal initial event state: " + e.getState());
```
arfathpasha: In IllegalArgumentException message, it would be good to add what the expected state is.

lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
```diff
@@ -0,0 +1,314 @@
...
+     * @param initialLoad the events to load into the queue.
+     */
+    public AccessGroupEventQueue(final Collection<StoredStatusEvent> initialLoad) {
+        Utils.nonNull(initialLoad, "initialLoad");
+        Utils.noNulls(initialLoad, "initialLoad has null entries");
+        final Map<String, StoredStatusEvent> objects = new HashMap<>();
+        final Map<String, List<StoredStatusEvent>> objectVersionReady = new HashMap<>();
+        final Map<String, List<StoredStatusEvent>> objectVersionProcessing = new HashMap<>();
+        
+        for (final StoredStatusEvent e: initialLoad) {
```
arfathpasha: coding style: In general I don't like using single letter variables like 'e' because it becomes hard to search through code for the variable. I tend to use acronyms 'sse' or just 'ee'. I don't expect you to go through and change all single letter vars as there are way too many to change, but just mentioning it as a good practice that can be followed in the future.

MrCreosote: My rules for single letter vars are that they need to be isolated to a < 10 line block or so, so it's trivial to see the entire scope of the variable. I think in that case it's ok.

lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
MrCreosote: Hmm... I assumed that the links in the documentation were enough. The error would alert the coder that they've passed in illegal events, and they could check the docs to see which events were ok.

arfathpasha: This is fine.

MrCreosote:
Merging with approval from @arfathpasha

PR #71, opened 2017-11-17
TASK-1160 add the overall event queue.

MrCreosote:


arfathpasha:
Need to add this to build.xml test target

<include name="kbasesearchengine/test/events/EventQueueTest.java"/>

MrCreosote:
@arfathpasha good catch, fixed

PR #82, opened 2017-11-30
TASK-1266 Remove some relation based cruft.

MrCreosote:
No plans to store relations in elastic at this point, simplify the code.

The primary key path is required - more on that later.

Also make the elastic search controller fail loudly.


arfathpasha:
There are some more references to "*relation*" in

- KBaseRelationEngine.spec
- scripts/run_perform.sh
- test/src/kbasesearchengine/test/KBaseRelationEngineServerTest.java
- build.xml

MrCreosote:
I've fixed these locally, but my local branch is pretty far ahead, so is it ok that the fixes go in a future PR?

PR #83, opened 2017-12-01
Remove unused booleans.

MrCreosote:
Code simplification

Note some of the changes are whitespace only.


MrCreosote:
Merging per @arfathpasha so we can get the coordinator and worker fix in. @arfathpasha will review later.

PR #85, opened 2017-12-02
TASK-1266 Reduce IndexerWorker API

MrCreosote:
many public methods that were not used publicly and don't need to be


MrCreosote:
@arfathpasha will review later

PR #86, opened 2017-12-02
Make coordinator and worker wait for task to stop

MrCreosote:
before letting the process die.


MrCreosote:
@arfathpasha will review later

PR #93, opened 2017-12-08
SCT-157 Refactor ObjectTypeParsingRules

MrCreosote:
* immutable
* fluent builder
* constraints enforced at build time rather than later in various places in the code
* defaults set at build time rather than calculated later in various places in the code
* no nulls returned from API
* split secondary utility code that builds OTPRs from various input sources into a separate class for cohesivity of the core OTPR value class
* rename methods to better reflect fuction

### DIFF
--- a/lib/src/kbasesearchengine/system/ObjectTypeParsingRulesUtils.java
+++ b/lib/src/kbasesearchengine/system/ObjectTypeParsingRulesUtils.java
@@ -1,0 +1,151 @@
+package kbasesearchengine.system;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+
+import kbasesearchengine.common.ObjectJsonPath;
+import kbasesearchengine.parse.ObjectParseException;
+import kbasesearchengine.system.ObjectTypeParsingRules.Builder;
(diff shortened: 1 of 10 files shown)


lib/src/kbasesearchengine/system/ObjectTypeParsingRulesUtils.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static ObjectTypeParsingRules fromStream(InputStream is, String sourceInfo) 
+            throws IOException, ObjectParseException {
+        @SuppressWarnings("unchecked")
+        Map<String, Object> obj = UObject.getMapper().readValue(is, Map.class);
+        return fromObject(obj, sourceInfo);
+    }
+
+    private static ObjectTypeParsingRules fromObject(
```
arfathpasha: The tests for this class have failure cases implemented but the exception lines in this method appear not to be covered in the jacoco output (KBaseSearchEngine/test_local/workdir/test-reports/html/kbasesearchengine.system/ObjectTypeParsingRulesUtils.java.html#L52)

MrCreosote: I intentionally didn't add any unit tests for this class because the pr was concentrating on the refactor of OTPR, not the utility method to load one from a file. It's no worse than before though. We'll add unit tests later.

PR #95, opened 2017-12-11
SCT-182 Object parser exception cleanup

arfathpasha:
A small PR for just the prepareTempFile function. I am working on one function at a time and making small PRs so I don't get mired in the parser rabbit hole. 

I also updated the doc pdf since I forgot to do it in the last update to docsource.


lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -476,10 +476,15 @@ private void indexObject(
             throws IndexingException, InterruptedException, RetriableIndexingException {
         long t1 = System.currentTimeMillis();
         final File tempFile;
+        File dir = getTempSubDir(guid.getStorageCode());
```
MrCreosote: Shouldn't this be inside the try block so SecurityExceptions from mkdirs() are caught?

lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -18,12 +18,25 @@
...
+     * temporary file.
+     */
+    public static File prepareTempFile(File tempDir)
+            throws IOException, SecurityException {
         if (!tempDir.exists()) {
             tempDir.mkdirs();
         }
-        File tempFile = File.createTempFile("ws_srv_response_", ".json", tempDir);
+        File tempFile = File.createTempFile(TEMP_FILE_PREFIX, TEMP_FILE_SUFFIX, tempDir);
+        tempFile.deleteOnExit();
```
MrCreosote: I would prefer to handle this by clearing the temp dir on startup so that temp files can be inspected if there's some error. The temp files are deleted in a finally block in the indexer worker.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -476,10 +476,15 @@ private void indexObject(
...
         try {
-            tempFile = ObjectParser.prepareTempFile(getTempSubDir(guid.getStorageCode()));
+            tempFile = ObjectParser.prepareTempFile(dir);
         } catch (IOException e) {
             throw new FatalRetriableIndexingException(e.getMessage(), e);
+        } catch (SecurityException e) {
+            logger.logError("Security Manager denied access to delete " +
```
MrCreosote: delete -> create?

lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -18,12 +18,25 @@
...
 public class ObjectParser {
-    
-    public static File prepareTempFile(File tempDir) throws IOException {
+    private static final String TEMP_FILE_PREFIX = "ws_srv_response_";
+    private static final String TEMP_FILE_SUFFIX = ".json";
+
+    /** Creates a temporary file in the specified directory. The file is deleted
+     * when the virtual machine terminates.
+     *
+     * @param tempDir
```
MrCreosote: missing description

arfathpasha:
I resorted to a better way to clean up the temp dir and temp file creation by factoring out that code from IndexerWorker and ObjectParser and placing it in a Util class.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -477,7 +478,8 @@ private void indexObject(
         long t1 = System.currentTimeMillis();
         final File tempFile;
         try {
-            tempFile = ObjectParser.prepareTempFile(getTempSubDir(guid.getStorageCode()));
+            FileUtil.getOrCreateCleanSubDir(rootTempDir, guid.getStorageCode());
```
MrCreosote: Deleting the subdir seems like a real problem here - firstly, it means that the workers can't be threaded, which is one of the items on the search priority list. Secondly, if I understand how things work with the indexing recursion, this will delete the files produced higher up in the recursion.

arfathpasha:
Good point. I reverted to just getOrCreate and no clean dir. I don't see a way to start with a clean temp dir that doesn't contain files from a previously failed execution (I noticed this can affect a future run some times and I have to manually clean the dir) as workers can be added at any point and there is no way to check if the worker being started is the first worker.

MrCreosote:
Also a good point. What if each worker has its own temp subdir with the worker name in the path, and we add a warning to SearchTools.java that all worker names must be unique?

There's probably some way of ensuring uniqueness but we can leave that for later.

arfathpasha:
How about we simply concat a "_"+java.util.UUID.randomUUID() to each worker temp subdir?

MrCreosote:
The reason I was thinking name is that it's easier to tell which directory goes to which worker process, and if no name is given the name is set to a UUID anyway. That gives the admin the choice as to whether to use UUIDs or names they determine.

MrCreosote:
i.e. https://github.com/kbase/KBaseSearchEngine/blob/master/lib/src/kbasesearchengine/tools/SearchTools.java#L275

arfathpasha:
Right. So I was thinking the two subDir names would look like - 
admin-id_UUD or UUID_UUID. The second UUID is like the "internally-generated characters" that the java.util.createTempFile() methods adds between the specified prefix and suffix. It's common for temp files and dirs to have this extra UUID. 

https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File-

MrCreosote:
Ok, so workername + uniqueness guarantee. Works for me.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -3,16 +3,7 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
MrCreosote: Pretty minor, but I don't like polluting the namespace with unnecessary names. It also makes the code somewhat harder to read if you're trying to figure out where a name came from and you're using a dumb text editor.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -82,13 +75,14 @@ public IndexerWorker(
             final IndexingStorage indexingStorage,
             final TypeStorage typeStorage,
             final File tempDir,
-            final LineLogger logger) {
+            final LineLogger logger) throws IOException {
```
MrCreosote: Very minor, but unless the method sig can fit on a single line IMO it's more readable if the exceptions get their own line. That separates them from the args.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -82,13 +75,14 @@ public IndexerWorker(
...
         Utils.notNullOrEmpty("id", "id cannot be null or the empty string");
         Utils.nonNull(logger, "logger");
         Utils.nonNull(indexingStorage, "indexingStorage");
         this.id = id;
         this.logger = logger;
-        this.rootTempDir = tempDir;
+        this.rootTempDir = FileUtil.getOrCreateCleanSubDir(tempDir,
+                id+"_"+ UUID.randomUUID().toString().substring(0,5));
```
MrCreosote: Should probably log the dir name so admins know which dirs are in use

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -104,29 +98,18 @@ public IndexerWorker(
             final IndexingStorage indexingStorage,
             final TypeStorage typeStorage,
             final File tempDir,
-            final LineLogger logger) {
+            final LineLogger logger) throws IOException {
```
MrCreosote: as above

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -104,29 +98,18 @@ public IndexerWorker(
...
         Utils.notNullOrEmpty("id", "id cannot be null or the empty string");
         Utils.nonNull(logger, "logger");
         this.id = id;
         this.storage = null;
-        this.rootTempDir = tempDir;
+        this.rootTempDir = FileUtil.getOrCreateCleanSubDir(tempDir,
+                id+"_"+ UUID.randomUUID().toString().substring(0,5));
```
MrCreosote: as above

test/src/kbasesearchengine/test/common/FileUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @AfterClass
+    public static void teardown() throws Exception {
+        FileUtils.deleteDirectory(tempDirPath.toFile());
+    }
+
+    @Test
+    public void getOrCreateSubDirTest() {
+        File subDir = FileUtil.getOrCreateSubDir(tempDirPath.toFile(), subDirName);
```
MrCreosote: Should test that the expected directory was returned

test/src/kbasesearchengine/test/common/FileUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertTrue("subdir was not created", subDir.isDirectory());
+    }
+
+    @Test
+    public void getOrCreateCleanSubDirTest() {
+        File subDir = FileUtil.getOrCreateSubDir(tempDirPath.toFile(), subDirName);
```
MrCreosote: As above

test/src/kbasesearchengine/test/common/FileUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @AfterClass
+    public static void teardown() throws Exception {
+        FileUtils.deleteDirectory(tempDirPath.toFile());
+    }
+
+    @Test
+    public void getOrCreateSubDirTest() {
+        File subDir = FileUtil.getOrCreateSubDir(tempDirPath.toFile(), subDirName);
+        assertTrue("subdir was not created", subDir.isDirectory());
```
MrCreosote: Ideally test the pre-existing directory case somehow

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: That's my intellij editor trying to be intelligent. I reverted.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Added info log.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Ditto.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

test/src/kbasesearchengine/test/common/FileUtilTest.java
MrCreosote: I meant that with this test, the fn could be returning any old directory and ignoring subDirName and tempDirPath.toFile(), and the test woudn't catch it. The test should check the contents of subDir to ensure they're as expected.

test/src/kbasesearchengine/test/common/FileUtilTest.java
MrCreosote: as above

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -75,14 +85,17 @@ public IndexerWorker(
...
+                throws IOException {
         Utils.notNullOrEmpty("id", "id cannot be null or the empty string");
         Utils.nonNull(logger, "logger");
         Utils.nonNull(indexingStorage, "indexingStorage");
         this.id = id;
         this.logger = logger;
         this.rootTempDir = FileUtil.getOrCreateCleanSubDir(tempDir,
                 id+"_"+ UUID.randomUUID().toString().substring(0,5));
+        logger.logInfo("Created temp dir "+rootTempDir.getAbsolutePath()+
+                                                     " for indexer worker "+id);
```
MrCreosote: Trivial, but IMO readability is better with spaces around operators.

MrCreosote:
Weird, a couple of the comments are shown as out of date although I just made them.

PR #98, opened 2017-12-14
SCT-178 Add a version int to search types

MrCreosote:
First step of making the search system aware of search type versions.

Lots still to do, and the version is ignored in most places.

* Update elastic search to take the version into account for indexes and
returned data
* Update TypeMappings to take the version of the type into account
* Update TypeFileStorage
* Add versions to type files (multiple versions per file), remove version key
* Add target type version to GUID transform

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -47,6 +47,7 @@ import kbasesearchengine.common.GUID;
 import kbasesearchengine.events.handler.SourceData;
 import kbasesearchengine.parse.ParsedObject;
 import kbasesearchengine.system.IndexingRules;
+import kbasesearchengine.system.SearchObjectType;
 import us.kbase.common.service.Tuple2;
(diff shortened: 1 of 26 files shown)


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1332,10 +1342,15 @@ public FoundHits searchObjects(String objectType, MatchFilter matchFilter,
...
-    private FoundHits queryHits(String objectType, List<Map<String, Object>> matchFilters, 
-            List<SortingRule> sorting, AccessFilter accessFilter, Pagination pg,
-            PostProcessing pp) throws IOException {
+    private FoundHits queryHits(
+            final SearchObjectType objectType,
+            final List<Map<String, Object>> matchFilters, 
+            List<SortingRule> sorting,
+            final AccessFilter accessFilter,
+            final Pagination pg,
+            final PostProcessing pp) throws IOException {
```
MrCreosote: Violated my own rule here. Fixed locally, will be in an upcoming PR.

PR #101, opened 2017-12-15
SCT-178 Add versions to search spec files

MrCreosote:
Files now support multiple versions per type in one file.

TODO:

* Need to add some tests here, but adding cases for incorrectly
formatted files is a real pain. Preferably add JSONSchema check for file
contents after a YAML parse.
* Make TypeMappings version aware.

### DIFF
--- a/resources/types/Genome.yaml
+++ b/resources/types/Genome.yaml
@@ -3,73 +3,76 @@
 
 # The name of the type in the search system.
 global-object-type: Genome
-global-object-type-version: 1
 # The data source of the objects, in this case the Workspace.
 storage-type: WS
 # The name of the type in the data source.
 storage-object-type: KBaseGenomes.Genome
 
-# The following rules describe which fields are extracted from the source object and transforms
-# performed upon said fields.
-indexing-rules:
+# This list contains the versions of the specification, ordered by version.
+versions:
     -
-        # Extract a string from the /id path. The key name will be id.
-        path: id
-        keyword-type: string
-    -
-        # Extract a string from the /domain path. The key name will be domain.
-        path: domain
-        keyword-type: string
-    -
-        # Extract a full-text search field from the /taxonomy path. A full-text field is parsed
-        # into tokens which can be searched on individually, as opposed to a string field which is
-        # an opaque block and must be matched in its entirety.
-        # The key name will be taxonomy.
-        path: taxonomy
-        full-text: true
-    -
-        # Extract a full-text search field from the /scientific_name path.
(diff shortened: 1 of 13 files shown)


resources/types/Genome.yaml
```diff
@@ -3,73 +3,76 @@
...
-global-object-type-version: 1
 # The data source of the objects, in this case the Workspace.
 storage-type: WS
 # The name of the type in the data source.
 storage-object-type: KBaseGenomes.Genome
 
-# The following rules describe which fields are extracted from the source object and transforms
-# performed upon said fields.
-indexing-rules:
+# This list contains the versions of the specification, ordered by version.
```
arfathpasha: This list contains the indexing rules for different versions of the storage-object-type specification, ordered by version number.

MrCreosote: This isn't quite right - the list can contain more than indexing rules, namely the subobject path, the subobject type, and the primary id path.

Also, the versions are versions of the search type, not the storage object type.

Happy to update the line with something that's more clear, but it needs to take the above into account.

arfathpasha: Yeah it is ambiguous. Would "parsing rules" be the right terminology that encapsulates indexing rules, subobject path, etc.

I understand the versions are the versions of the search type, but aren't these versions mapped one-to-one to the storage-object-type versions? If so, stating this will make it a little more clear.

MrCreosote: "Parsing rules" sounds good.

No, the search versions and storage versions are not directly coupled. This is because a) a storage version change does not necessarily require a search version change and b) we may want to make changes to the search version (to add, change, or remove fields) when there's no change on the storage side. The type mappings establish the mapping from version to version.

arfathpasha: Thanks for clarifying. 

Is this description correct?

This list contains different versions of the parsing rules used to parse the specified storage-object-type specification. The list is ordered by version number. Each parsing rule version results in a corresponding search type version. The mapping between the source storage-object-type version and the target search type version is specified by the typemappings.

MrCreosote: With the addition that if no typemappings are available for a type the last version is used.

arfathpasha: OK. This additional caveat I think should go into the typemappings spec.

MrCreosote: Hm, I'm thinking both places. You def want to know which search version is going to be used in the context of the search spec. 

This'll be in the next PR.

PR #105, opened 2017-12-18
SCT-182: ObjectParser documentation

arfathpasha:


lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -80,7 +121,8 @@ public static String extractParentFragment(ObjectTypeParsingRules parsingRules,
             if (!rules.isFromParent()) {
                 continue;
             }
-            //TODO CODE this seems wrong. Why adding null paths?
+            // This list add is simply a check to see if at least one rule for
+            // extracting a parent element exists.
```
MrCreosote: I'm not sure this is true - the indexing paths are used in line 92/134 below.

lib/src/kbasesearchengine/parse/KeywordParser.java
```diff
@@ -254,31 +254,27 @@ private static Object transform(Object value, IndexingRules rule,
...
             final String storageCode = typeDescr.getStorageObjectType().getStorageCode();
             final Set<String> refs = toStringSet(value);
             final Set<GUID> unresolvedGUIDs;
-            try {
-                unresolvedGUIDs = refs.stream().map(r -> GUID.fromRef(storageCode, r))
+            unresolvedGUIDs = refs.stream().map(r -> GUID.fromRef(storageCode, r))
                         .collect(Collectors.toSet());
-            } catch (IllegalArgumentException e) {
-                throw new ObjectParseException(e.getMessage(), e);
-            }
```
MrCreosote: in this case the illegal argument exception is thrown because the reference extracted from the input object is illegal, so an objectparseexception is the right exception to throw.

lib/src/kbasesearchengine/common/JsonTokenUtil.java
```diff
@@ -161,19 +162,15 @@ public static Object getCurrentTokenPrimitive(JsonParser jts,
...
             return true;
         } else {
-            throw new IOException("Unexpected token type: " + t);
+            throw new ObjectParseException("Unexpected token type: " + t);
         }
     }
 
-    public static String prettyPrint(Object obj) {
-        try {
+    public static String prettyPrint(Object obj) throws IOException {
```
MrCreosote: Is this code even used anywhere? Delete maybe?

lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -57,21 +72,47 @@
...
+     *
+     *
+     * @param parsingRules parsing rules
+     * @param guid GUID without the sub-object path and id
+     * @param path
+     * @param idConsumer
+     * @return GUID with sub-object path and id if data needs to be extracted
+     * from sub-object, lese GUID with just main object information.
+     */
+    public static GUID prepareGUID(
```
MrCreosote: Out of scope, but I really dislike that this method is public. We should (later) try to make it private and remove the id-defaulting-to-path stuff.

lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -57,21 +72,47 @@
...
-        return new GUID(guid, innerSubType, innerID);
+        return new GUID(guid, subObjectType, subObjectID);
     }
-    
-    public static String extractParentFragment(ObjectTypeParsingRules parsingRules,
-            JsonParser jts) throws ObjectParseException, IOException {
+
+    /** Returns null if the specified parsing rules do not have rules for
+     * extracting parent elements. Else returns "/" for the root path of the
+     * parent json object.
```
MrCreosote: I don't think this is true - I believe what's returned is a JSON string containing the extracted portions of the parent object.

lib/src/kbasesearchengine/parse/ObjectParser.java
MrCreosote: Actually I think that's true about virtually every method in this class. I'd prefer to have one API call that takes the source data (+ whatever else is needed) and returns a single value object containing the resulting extracted data. I think all this stuff is batched anyway...

lib/src/kbasesearchengine/common/JsonTokenUtil.java
arfathpasha: Removed.

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: Reverted

lib/src/kbasesearchengine/parse/ObjectParser.java
arfathpasha: I totally agree. If we reused an external package like JsonSurfer, we would end up with a single API call. I will leave this as-is for now and will make a note to come back to this post contractor meeting and re-evaluate the option to reuse external libs here and simplify the code that way.

lib/src/kbasesearchengine/parse/ObjectParser.java
arfathpasha: You are right! I overlooked the nested calls to pathToJson. Documentation has been fixed.

lib/src/kbasesearchengine/parse/ObjectParser.java
arfathpasha: Yep. Removed comment.

lib/src/kbasesearchengine/common/JsonTokenUtil.java
```diff
@@ -64,7 +65,7 @@ public static ObjectJsonPath trimPath(ObjectJsonPath jsonPath) {
      * we read field token before processing value block.
      */
     public static void writeTokensFromCurrent(final JsonParser jts, final JsonToken current,
-            final JsonGenerator jgen) throws IOException {
+            final JsonGenerator jgen) throws ObjectParseException, IOException {
```
MrCreosote: Is IOExeption thrown from this method anymore? If not, remove. Same question for the other methods that throw IOException.

lib/src/kbasesearchengine/parse/ObjectParser.java
```diff
@@ -80,7 +121,7 @@ public static String extractParentFragment(ObjectTypeParsingRules parsingRules,
             if (!rules.isFromParent()) {
                 continue;
             }
-            //TODO CODE this seems wrong. Why adding null paths?
+
```
MrCreosote: Could you restore this TODO? Adding nulls to this list still seems like the wrong thing to do to me, and if it ever actually happens I expect NPEs to get thrown.

arfathpasha: Done. I'll fix this and test in the next PR.

lib/src/kbasesearchengine/common/JsonTokenUtil.java
arfathpasha: Yes it is being thrown by JsonGenerator.writeStartArray()

PR #108, opened 2017-12-20
[KBASE-4234]when search term is multiple words, default should be AND, not OR	

tiramisu24:
-- API change of elasticsearch query call from ors to ands
-- Add test

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -39,16 +39,14 @@ import org.elasticsearch.client.Response;
 import org.elasticsearch.client.ResponseException;
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestClientBuilder;
-
-import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableMap;
 
 import kbasesearchengine.common.GUID;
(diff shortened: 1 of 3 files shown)


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -64,16 +62,13 @@
...
     private static final String OBJ_NAME = "oname";
-    
-    private static final String SEARCH_OBJ_TYPE = "otype";
-    private static final String SEARCH_OBJ_TYPE_VER = "otypever";
 
     private HttpHost esHost;
     private String esUser;
     private String esPassword;
     private String indexNamePrefix;
+    private boolean mergeTypes = false;
```
MrCreosote: There's something very wrong here. I specifically removed this field in an earlier PR and your PR is putting it back - there must be something funky in your fork.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -135,67 +147,35 @@ public void dropData() throws IOException {
         }
     }
     
-    /* checks that at least one index exists for a type, and throws an exception otherwise.
```
MrCreosote: Same here. I added these functions in a recent PR.

PR #109, opened 2017-12-21
[KBASE-4234] when search term is multiple words, default should be AND, not OR	

tiramisu24:
-api change for multiword query to elasticsearch to default to AND instead of OR 
-add test


lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -295,6 +295,37 @@ public SearchObjectsOutput searchObjects(SearchObjectsInput params, String user)
         ret.withSearchTime(System.currentTimeMillis() - t1);
         return ret;
     }
+     //for testing purposes. Access set to admin
```
MrCreosote: C&Ping this method to merely add admin creds seems like a bad idea. Firstly, if we really need an admin method, the right thing to do is modify the source method to take an asAdmin boolean or something like that and set up the access filter based on said boolean. However, I don't really understand why that's needed at all - is it not possible to call the method with proper input so the user gets treated as an admin?

Further though - all the changes are in the ElasticIndexingStorage class so why not just write tests against that class directly instead of indirectly via SearchMethods?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1267,7 +1276,17 @@ public FoundHits searchObjects(String objectType, MatchFilter matchFilter,
     private List<Map<String, Object>> prepareMatchFilters(MatchFilter matchFilter) {
         List<Map<String, Object>> ret = new ArrayList<>();
         if (matchFilter.fullTextInAll != null) {
-            ret.add(createFilter("match", "_all", matchFilter.fullTextInAll));
+            LinkedHashMap<String, Object> query = new LinkedHashMap<String, Object>() {{
```
MrCreosote: Is createFilter used anywhere else, and so should the changes go into createFilter rather than bypassing it? Is there anything in createFilter that the new code is not doing but should be?

tiramisu24: b/c createFilters expects everything to be singly nested, something like <a :{b:c}> and I want something like <a:{b:c, d}>

lib/src/kbasesearchengine/main/SearchMethods.java
tiramisu24: I actually started writing it in elasticindexingstorage but I needed the same classes but didn't need to invoke all the parameters. so I ended up moving things out and making mocks. And since I was doing that already, it was easier to call on SearchMethods directly so that I didn't have to copy most of the SearchObjectsOutput method in the setup. And yeah I ended up having to copy the method b/c of the accessFilter is kind of messed up. It takes a username and changes it from kbaseengine.AccessFilter to kbaseengine.search.AccessFilter <= this one can be set as admin. I could move it back in?

MrCreosote: I don't think we should be copying searchObjects anywhere - we should be doing unit tests against EIS similar to the other unit tests.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: The question is whether create filter should be changed to the new behavior rather than bypassing it.

tiramisu24: maybe. Its not being consistently used in the rest of the methods either.

tiramisu24: it may require a bunch of small changes everywhere though...

PR #110, opened 2017-12-22
SCT-285 ElasticIndexingStorage cleanup

arfathpasha:
- removed double curly brace antipattern as it causes a lot of load on the classloader. https://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/
- added some documentation around the creation of the ES requests

- squeezed in an update to the high level design


MrCreosote:
> removed double curly brace antipattern 

I love you for this

docsource/architectureHighLevelDesign.rst
```diff
@@ -20,8 +20,10 @@ Data Flow
```
MrCreosote: If you redo the diagram again, strictly speaking the search api is a part of KBaseSearchEngine. I think there's 3 main parts:

The indexer - includes mongo, coodinator, workers
Data storage - elastic
search API

Not holding up the PR for that though.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -331,21 +331,26 @@ public void flushIndexing(SearchObjectType objectType) throws IOException {
```
MrCreosote: Things would probably be even more readable if we used ImmutableMap.of() calls to build the maps in a nested fashion rather than declaring each map as a top level variable. I wish java had map and list literals but ImmutableMap is the closest I've found so far. Again, not blocking this PR - getting rid of the anon classes is great.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -664,14 +702,14 @@ private boolean updateAccessGroupForVersions(
...
-        }};
-        Map<String, Object> doc = new LinkedHashMap<String, Object>() {{
-            put("query", query);
-            put("script", script);
-        }};
+        Map<String, Object> script = new LinkedHashMap();
+        script.put("inline", inline.toString());
+        script.put("params", params);
+
+        Map<String, Object> doc = new LinkedHashMap<String, Object>();
```
MrCreosote: No need for the RHS generic type declaration, not blocking.

MrCreosote: Few more below

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -689,32 +727,32 @@ private boolean removeAccessGroupForVersion(String indexName, GUID guid,
...
-        }};
+        Map<String, Object> script = new LinkedHashMap();
+        script.put("inline", "" +
+                "ctx._source.lastin.remove(ctx._source.lastin.indexOf(params.accgrp));\n" +
+                "if (ctx._source.extpub.indexOf(params.accgrp) >= 0) {\n" +
+                "  ctx._source.extpub.remove(ctx._source.extpub.indexOf(params.accgrp));\n" +
+                "}\n" +
+                (fromAllGroups ? (
+                        "int pos = ctx._source.groups.indexOf(params.accgrp);\n" +
+                                "if (pos >= 0) {\n" +
```
MrCreosote: Don't think this should be indented

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1624,71 +1723,99 @@ private void createTables(String indexName, List<IndexingRules> indexingRules) t
...
-            put("type", "keyword");
-        }});
+
+        tmp = new LinkedHashMap();
+        tmp.put("type", "date");
+        props.put("timestamp", tmp);
+
+        tmp = new LinkedHashMap();
+        tmp.put("type", "keyword");
+        props.put("prefix", tmp);
```
MrCreosote: Could replace tmp with the keyword variable here

PR #111, opened 2018-01-02
Clean up warnings

MrCreosote:
Warnings have been accumulating in the codebase; this commit fixes all
that Eclipse detects.

Also makes warning suppression more specific - adds suppression to
variables rather than methods.


arfathpasha:
I commented out the @SuppressWarning annotations in ElasticIndexingStorage.java and ran tests and didn't see compiler throw any warnings. I wonder if eclipse if being over paranoid.

PR #114, opened 2018-01-04
[KBASE-4234] when search term is multiple words, default should be AND, not OR	#109

tiramisu24:
-api change for multiword query to elasticsearch to default to AND instead of OR
-add test
-add test for searching mulitiple key filters


.gitignore
```diff
@@ -6,3 +6,5 @@ bin
 /search_tools.cfg
 /classes/
 /search_tools.cfg.bak
+.idea/
+test/ .idea/
```
MrCreosote: The test directory cannot be ignored.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1320,7 +1325,16 @@ public FoundHits searchObjects(String objectType, MatchFilter matchFilter,
     private List<Map<String, Object>> prepareMatchFilters(MatchFilter matchFilter) {
         List<Map<String, Object>> ret = new ArrayList<>();
         if (matchFilter.fullTextInAll != null) {
-            ret.add(createFilter("match", "_all", matchFilter.fullTextInAll));
```
MrCreosote: Just to be clear, because I still don't understand - should the create filter method be modified with this new behavior?

MrCreosote: Also, do all queries now use AND logic rather than OR, and so have consistent behavior?

MrCreosote: Or should createFilter be modified to support both the old and new behavior, controlled by a boolean?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -3,6 +3,7 @@
 import static kbasesearchengine.test.common.TestCommon.set;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
```
MrCreosote: don't use assertTrue, use assertThat("message", got, is(true));

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -19,11 +20,10 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import kbasesearchengine.search.*;
```
MrCreosote: Please don't use wildcard imports.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -19,11 +20,10 @@
...
 import org.apache.commons.io.FileUtils;
 import org.apache.http.HttpHost;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.*;
```
MrCreosote: wildcard imports again

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -43,19 +43,14 @@
...
-import kbasesearchengine.search.ObjectData;
-import kbasesearchengine.search.PostProcessing;
 import kbasesearchengine.system.IndexingRules;
 import kbasesearchengine.system.ObjectTypeParsingRules;
 import kbasesearchengine.system.ObjectTypeParsingRulesFileParser;
 import kbasesearchengine.system.SearchObjectType;
 import kbasesearchengine.test.common.TestCommon;
 import kbasesearchengine.test.controllers.elasticsearch.ElasticSearchController;
 import kbasesearchengine.test.parse.SubObjectExtractorTest;
+import org.mockito.Mock;
```
MrCreosote: This appears to be unused

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -619,6 +614,154 @@ public void testTypeVersions() throws Exception {
...
         
+    }
+
+    private void prepareTestMultiwordSearch() throws ObjectParseException {
+        SearchObjectType objectType = new SearchObjectType("Simple", 1);
+        IndexingRules ir = IndexingRules.fromPath(new ObjectJsonPath("prop1"))
+                .withFullText().build();
+        List<IndexingRules> indexingRules= Arrays.asList(ir);
+        try {
+            indexObject(new GUID("WS:11/1/2"), objectType, "{\"prop1\":\"multiWordInSearchMethod1 multiWordInSearchMethod2\"}", "multiword.1", Instant.ofEpochSecond(1514939000), null,
```
MrCreosote: 100 chars per line max

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -619,6 +614,154 @@ public void testTypeVersions() throws Exception {
...
+    private void prepareDatabase() throws ObjectParseException {
+        SearchObjectType objType = new SearchObjectType("SimpleNumber",1 );
+        IndexingRules ir1 = IndexingRules.fromPath(new ObjectJsonPath("num1"))
+                .withKeywordType("integer").build();
+        IndexingRules ir2 = IndexingRules.fromPath(new ObjectJsonPath("num2"))
+                .withKeywordType("integer").build();
+        List<IndexingRules> indexingRules= Arrays.asList(ir1,ir2);
+
+        try {
+            indexObject(new GUID("WS:12/1/1"), objType, "{\"num1\": 123, \"num2\": 123}", "number.1", Instant.now(), null,
```
MrCreosote: 100 chars

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: my impression was that the change was to get ride of the old behavior entirely

MrCreosote: So the createFilter method should be removed then?

tiramisu24: yes. Most methods that generates a call to elasticsearch calls prepareMatchFilters. The exceptions are the indexing methods (which I don't need to change) and getObjectsByIds. I think that is the only place that should have the old logic (OR) b/c of the nature of that function which allows you to pass in multiple ids....

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: sure. is assertThat better?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: ughh I think thats due to intellij auto formatting. need to figure out how to turn that off

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: actually createFilter should be renamed (not sure what maybe HelpMeNestThingsSingularly). Its used everywhere as helper to nest commands. For ex. updateAccessGroupForVersions, setFieldOnObject and a bunch of methods use it. These methods make post request and so is not what I am looking to change.... 

None of the search methods directly call it (only through prepareMatchFilters) and the fullTextInAll field is the only place I need to double nest things. objectName don't have spaces and lookupInKeys uses another type of query that defaults to "AND" behavior already.

MrCreosote: So now what I'm hearing is that createFilter's behavior should stay as is (with a possible fn rename to better state what the fn does) and the new behavior you're adding is only relevant in the one place it currently exists?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Basically assert* are all just special cases of assertThat, so by using assertThat you have consistent assertions.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -2,7 +2,9 @@
 
 import static kbasesearchengine.test.common.TestCommon.set;
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.equalTo;
```
MrCreosote: Just use is - equivalent behavior.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+        assertThat("did not find object1 using LookupInKey", hits2.guids, equalTo(set(guid2, guid3)));
+        assertThat("did not find object1 using LookupInKey", hits3.guids, equalTo(set(guid1, guid2)));
+
+        //conflicting filters should return nothing
+        MatchFilter filter4 =  MatchFilter.create().withLookupInKey("num1", range1);
+        filter4.withLookupInKey("num2", range2);
+        FoundHits hits4 = indexStorage.searchObjects(null, filter4,sorting, accessFilter
+                , null, null);
+
+        assertThat("conflicting ranges should produce 0 results", hits4.guids.size() == 0,
```
MrCreosote: isEmpty() is cleaner

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+                    true, indexingRules);
+            indexObject(guid3, objectType, "{\"prop1\":\"multiWordInSearchMethod1\"}",
+                    "multiword.3", Instant.now(), null,
+                    true, indexingRules);
+    }
+
+
+    @Test
+    public void testMultiwordSearch() throws Exception{
+        GUID guid1 = new GUID("WS:11/1/2");
```
MrCreosote: Out of curiosity why are you using version 2? Doesn't matter, just wondering.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+        FoundHits hits2 = indexStorage.searchObjects(null, filter,sorting, accessFilter
+                , null, null);
+
+
+        filter.withFullTextInAll("multiWordInSearchMethod2");
+        FoundHits hits3 = indexStorage.searchObjects(null, filter,sorting, accessFilter
+                , null, null);
+
+        assertThat("did not find object1", hits1.guids, equalTo(set(guid1)));
+        assertThat("did not find object2", hits2.guids, equalTo(set(guid1,guid3)));
```
MrCreosote: Message doesn't match test

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+
+        FoundHits hits1 = indexStorage.searchObjects(null, filter1,sorting, accessFilter
+                , null, null);
+        FoundHits hits2 = indexStorage.searchObjects(null, filter2,sorting, accessFilter
+                , null, null);
+        FoundHits hits3 = indexStorage.searchObjects(null, filter3,sorting, accessFilter
+                , null, null);
+
+        assertThat("did not find object1 using LookupInKey", hits1.guids, equalTo(set(guid1)));
+        assertThat("did not find object1 using LookupInKey", hits2.guids, equalTo(set(guid2, guid3)));
```
MrCreosote: message doesn't match test

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+
+        prepareTestMultiwordSearch(guid1, guid2, guid3);
+        final kbasesearchengine.search.MatchFilter filter = new kbasesearchengine.search.MatchFilter();
+
+        List<kbasesearchengine.search.SortingRule> sorting = null;
+        AccessFilter accessFilter = AccessFilter.create().withAdmin(true);
+
+        filter.withFullTextInAll("multiWordInSearchMethod1 multiWordInSearchMethod2");
+        FoundHits hits1 = indexStorage.searchObjects(null, filter,sorting, accessFilter
+                , null, null);
```
MrCreosote: Commas should go on the prior line

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
         
     }
 
+    private void prepareTestMultiwordSearch(GUID guid1, GUID guid2, GUID guid3) throws Exception {
+        SearchObjectType objectType = new SearchObjectType("Simple", 1);
+        IndexingRules ir = IndexingRules.fromPath(new ObjectJsonPath("prop1"))
+                .withFullText().build();
+        List<IndexingRules> indexingRules= Arrays.asList(ir);
```
MrCreosote: spaces around operators and  =
more below

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+        indexObject(guid3, objType, "{\"num1\": 1236, \"num2\": 1236}",
+                "number.3", Instant.now(), null,
+                false, indexingRules);
+    }
+    @Test
+    public void testLookupInKey() throws Exception{
+        GUID guid1 = new GUID("WS:12/1/2");
+        GUID guid2 = new GUID("WS:12/2/2");
+        GUID guid3 = new GUID("WS:12/3/2");
+        prepareTestLookupInKey(guid1,guid2,guid3);
```
MrCreosote: space after comma

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -621,4 +625,129 @@ public void testTypeVersions() throws Exception {
...
+                , null, null);
+        assertThat("did not find object1 using LookupInKey", hits0.guids, equalTo(set(guid1)));
+
+
+        //key, range lookup
+        MatchValue range1 = new MatchValue(100, 200);
+        MatchValue range2 = new MatchValue(1000, 2000);
+        MatchValue range3 = new MatchValue(100, 1234);
+
+        MatchFilter filter1 =  MatchFilter.create().withLookupInKey("num1", range1);
```
MrCreosote: extra space after =

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: yes. It actually makes more sense to rename fn prepareMatchfilters as "prepareFilter"(or "createFilters") s cuz that is what it is doing. Esp since elasticsearch call their filters filters and not matchfilters... The current fn createFilters is more like add all kinds of key/value pairs, not just filters, to calls

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: b/c when I was starting to write the tests I was curious if it the database would let me. Got distracted with other things and forgot that I left it as version 2....

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: ops I should not copy and paste

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -73,6 +73,14 @@ private ObjectData(
     /** Get the object's GUID.
      * @return the GUID.
      */
+    public String toString() {
```
MrCreosote: This either needs tests or needs to be removed

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -21,12 +21,20 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import kbasesearchengine.search.AccessFilter;
```
MrCreosote: is there a reason these were moved way up here?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -591,11 +593,14 @@ public void testTypeVersions() throws Exception {
                 .withNullableObjectName("o1")
                 .withNullableType(type1)
                 .withNullableCreator("creator")
-                .withNullableTimestamp(now)
+                .withNullableTimestamp(indexedObj1.getTimestamp().get())
                 .withNullableData(ImmutableMap.of("bar", 1))
                 .withKeyProperty("bar", "1")
                 .build();
         
+        //due to potential truncation of timestamp on mac
```
MrCreosote: Need this same line for indexdObj2

MrCreosote:
Also I strongly recommend merging master so that the automated tests and coverage work.

MrCreosote:
Per steve, go ahead and merge master

tiramisu24:
alright >__>

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: intellij auto format

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: done

tiramisu24: done

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Can you move them back where they belong please?

tiramisu24: sure but it will still show changed b/c I have to manually move them

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -73,6 +73,7 @@ private ObjectData(
     /** Get the object's GUID.
      * @return the GUID.
      */
+
```
MrCreosote: Might as well remove the space here

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -45,19 +47,20 @@
...
-import kbasesearchengine.search.ObjectData;
-import kbasesearchengine.search.PostProcessing;
 import kbasesearchengine.system.IndexingRules;
 import kbasesearchengine.system.ObjectTypeParsingRules;
 import kbasesearchengine.system.ObjectTypeParsingRulesFileParser;
 import kbasesearchengine.system.SearchObjectType;
 import kbasesearchengine.test.common.TestCommon;
 import kbasesearchengine.test.controllers.elasticsearch.ElasticSearchController;
 import kbasesearchengine.test.parse.SubObjectExtractorTest;
+import kbasesearchengine.search.AccessFilter;
```
MrCreosote: Can you put these back where they belong please?

PR #115, opened 2018-01-04
Handle headless indexer & worker, attempt 1

MrCreosote:
No console if run headless.

Headless mode needs testing.

### DIFF
--- a/lib/src/kbasesearchengine/tools/SearchTools.java
+++ b/lib/src/kbasesearchengine/tools/SearchTools.java
@@ -108,7 +108,6 @@ public class SearchTools {
         nonNull(args, "args");
         nonNull(out, "out");
         nonNull(err, "err");
-        nonNull(console, "console");
         this.args = args;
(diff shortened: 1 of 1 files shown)


lib/src/kbasesearchengine/tools/SearchTools.java
```diff
@@ -223,10 +223,27 @@ public int execute() {
     }
     
     private void waitForReturn(final Stoppable stoppable) throws InterruptedException {
-        out.println("Press return to shut down process.");
-        console.readLine();
-        System.out.println("Waiting for task to stop");
-        stoppable.stop(10 * 60 * 1000); // 10 mins
+        if (console == null) {
```
sychan: Don't you also need to deal with the nonNull() call on line 111 so that it doesn't abort at startup?

MrCreosote: yep, forgot. fixed

PR #116, opened 2018-01-05
SCT-178 Add another TypeFileStorage test; TFS tests part 2/4

MrCreosote:


MrCreosote:
@ugswork 
I don't quite understand your first comment. What names are you talking about? Variable names in the tests?

For the 2nd comment, the resettable input stream is only needed when the stream is read more than once during the test. If that happens for a normal stream, the test will fail. The only reason to use the resettable stream for the tests you mention is consistency, which is why I didn't bother to go back and change the pre-existing tests.

ugswork:
@MrCreosote,

>>> don't quite understand your first comment. What names are you talking about? Variable names >>> in the tests?

I meant the following:
instead of:     StorageObjectType [storageCode=CD, type=storefoo]
we may use:  StorageObjectType [storageCode=WS, type=KBaseGenomeAnnotations.Assembly]
instead  of:    [SearchObjectType [type=foo, version=1]]
                or . [SearchObjectType [type=bar, version=1]]
we may use:  [SearchObjectType [type=Assembly, version=1]]
                or   [SearchObjectType [type=AssemblyContig, version=1]]

>>> For the 2nd comment, the resettable input stream is only needed when the stream is read more >>> than once during the test. If that happens for a normal stream, the test will fail. The only reason >>> to use the resettable stream for the tests you mention is consistency, which is why I didn't 
>>> bother to go back and change the pre-existing tests.

I meant only for consistency. So no need to do it.

MrCreosote:
I see. Maybe when the tests are all merged in you could make a PR for the changes you suggest?

PR #122, opened 2018-01-09
SCT-285 - Allow mulitple types to search on

arfathpasha:
Updated spec, implementation and added a test.


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1388,8 +1388,22 @@ private FoundHits queryHits(
...
-        String indexName = objectType == null ? getAnyIndexPattern() : checkIndex(objectType);
-        String urlPath = "/" + indexName + "/" + getDataTableName() + "/_search";
+
+        StringBuffer indexName = new StringBuffer();
+        for(int ii=0; ii<objectTypes.size(); ii++) {
+
+            String objectType = objectTypes.get(ii);
+
+            indexName.append(
+                    objectType == null ? getAnyIndexPattern() : checkIndex(objectType)
```
MrCreosote: I don't get this. If there's a null entry in the list of types, it adds a search term that matches all types to the url? That means if there's a null in the list, there's no point in doing anything other than making the url the result of getAnyIndexPattern() since that's a completely open search re types.

It makes more sense to me to throw an exception if there are any null entries in the type list (Utils.noNulls()), and if the type list is null or empty then the url is getAnyIndexPattern(), otherwise the url is constructed from the type strings via checkIndex().

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1388,8 +1388,22 @@ private FoundHits queryHits(
...
+            );
+
+            if( ii < objectTypes.size()-1 )
```
MrCreosote: Simpler implementation would be something like

String indexName = String.join(",", objectTypes.stream().map(s -> checkIndex(s)).collect(Collectors.toList()));

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1388,8 +1388,22 @@ private FoundHits queryHits(
...
+                indexName.append(",");
+        }
+
+
+        String urlPath = "/" + indexName.toString() + "/" + getDataTableName() + "/_search";
```
MrCreosote: how many types in one call can we support given URL length limitations? Should we throw an error if objectTypes.size() > MAX_TYPES_SUPPORTED

MrCreosote: Alternatively is there a way to specify the indexes to be searched without putting it in the url?

arfathpasha: I don't think ES has a way of specifying the multiple types in the body of the message. It also looks like ES does not limit the length of the URL although it is configurable and the default is 4k bytes. 

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java

I have added a check for this default size.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: I tried this but it got all convoluted because checkIndex throws an exception and the syntax for handling that exception looked real messy. 

I am not that adept with lamba expressions in Java yet. Perhaps I should spend one of the weekly 10% time on this.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: I figured that if there was a null in the list, it would just end up relaxing the search to an unconstrained search. Anyhow I have added object type validation.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Lots of discussion around this here. Some info I've found so far:

https://github.com/elastic/elasticsearch/issues/26360
https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-multi-search.html

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Most of the benefit comes from the String.join() IMO. Lambdas and checked exceptions don't play well so try:

final List<String> r = new LinkedList<>();
for (final String type: objetTypes) {
    r.add(checkIndex(type));
}
final String indexName = String.join(",", r);

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Index blacklists are possible: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/multi-index.html

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: This API still seems weird to me. If I want to search across all types, it seems intuitive to me to just not include a type list at all or pass an empty list. Passing a list with a single null element is not what I would expect at all.

arfathpasha: Yes it is weird. I was trying to make minimum changes to get this out fast. It is now fixed. Empty list implies anyType. null in list in invalid.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Done.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -7,18 +7,7 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
```
MrCreosote: Please don't use wildcard imports

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +125,41 @@ public void dropData() throws IOException {
...
     }
-    
-    /* checks that at least one index exists for a type, and throws an exception otherwise.
-     * Returns a index string that matches any version of the type.
+
+
+    /** The specified list is valid if it,
+     *
+     * 1. is empty,
+     * 2. or one or more non-null elements.
```
MrCreosote: contains

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +125,41 @@ public void dropData() throws IOException {
...
         }
     }
-    
-    /* checks that at least one index exists for a type, and throws an exception otherwise.
-     * Returns a index string that matches any version of the type.
+
+
+    /** The specified list is valid if it,
+     *
+     * 1. is empty,
```
MrCreosote: I'm fine with the ElasticIndexingStorage level rejecting null search lists, but I do think that leaving the object_types field out of the incoming json should result in a non-type-constrained search rather than an illegal argument exception. If that behavior is controlled in a layer above EIS that's ok with me.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +125,41 @@ public void dropData() throws IOException {
...
+
+        if(objectTypes== null) {
+            throw new IllegalArgumentException("Invalid list of object types. List is null.");
+        }
+
+        if(objectTypes.isEmpty()) {
+            return;
+        }
+
+        if(objectTypes.size() > 0 && objectTypes.contains(null)) {
```
MrCreosote: size() > 0 is guaranteed here

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed here. I need to fix this in intellij too!

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: I have added a bit of validation to ensure that an empty list is created if it is left out in the request from the client.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: To record what we said in the standup:
1) Add a maximum search type size in SearchObjectType (say 50 bytes of utf-8)
2) Add a maximum count of search types to search (say 50) on and throw and exception if it's larger than that number

search type size * search type count << 4096 so we have slop for doing blacklists, preventing sub type searches, etc

Blacklists and preventing sub type searches we'll do in another PR.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Removed.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Completed.

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -267,6 +268,12 @@ public SearchTypesOutput searchTypes(SearchTypesInput params, String user) throw
     public SearchObjectsOutput searchObjects(SearchObjectsInput params, String user) 
             throws Exception {
         long t1 = System.currentTimeMillis();
+
```
MrCreosote: We really need to write tests for SearchMethods at some point, but that's not blocking this PR IMO

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: this is weird. I see that you fixed this in one of the commits, but in the overall PR it still shows it as a wildcard import.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -81,6 +70,16 @@
     public static final int PUBLIC_ACCESS_GROUP = -1;
     public static final int ADMIN_ACCESS_GROUP = -2;
 
+    /** ElasticSearch max length of an HTTP URL. Defaults to 4kb
```
MrCreosote: I think the first line here is out of date

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +133,47 @@ public void dropData() throws IOException {
...
+     *
+     * 1. is empty,
+     * 2. or contains one or more non-null elements and size is less than
+     *    {@link ElasticIndexingStorage#MAX_OBJECT_TYPES_SIZE}.
+     *
+     * list 1 represents a list that would map to any index pattern (search unconstrained by type)
+     * list 2 represents a list that would map to one or more specific index patterns (constrained search)
+     *
+     *
+     * @param objectTypes a non-empty list of object types.
```
MrCreosote: strike non-empty

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +133,47 @@ public void dropData() throws IOException {
...
+     * @throws IOException if list is invalid.
+     */
+    private void validateObjectTypes(List<String> objectTypes) throws IOException {
+
+        if(objectTypes== null) {
```
MrCreosote: Can you leave a space after if to show that it's not a function?

MrCreosote: Also spaces around operators, == in this case

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -134,9 +133,47 @@ public void dropData() throws IOException {
...
+            throw new IllegalArgumentException("Invalid list of object types. List is null.");
+        }
+
+        if(objectTypes.isEmpty()) {
+            return;
+        }
+
+        if(objectTypes.size() > MAX_OBJECT_TYPES_SIZE) {
+            throw new IOException("Invalid list of object types. " +
+                    "List size exceeds maximum limit of "+MAX_OBJECT_TYPES_SIZE);
```
MrCreosote: Spaces around operators

lib/src/kbasesearchengine/system/SearchObjectType.java
```diff
@@ -12,20 +13,30 @@
...
      * less than one.
      */
     public SearchObjectType(final String type, final int version) {
         Utils.notNullOrEmpty(type, "search type cannot be null or whitespace");
+
+        try {
+            if( type.getBytes("UTF-8").length > MAX_TYPE_SIZE) {
+                throw new IllegalArgumentException("Search type string size must be less than "+MAX_TYPE_SIZE+" UTF-8 chars");
+            }
+        } catch(UnsupportedEncodingException ex ) {
```
MrCreosote: space after catch to show it's not a function

lib/src/kbasesearchengine/system/SearchObjectType.java
```diff
@@ -12,20 +13,30 @@
...
      * @param version the version of the type.
      * @throws IllegalArgumentException if the type name is null or whitespace or the version is
      * less than one.
      */
     public SearchObjectType(final String type, final int version) {
         Utils.notNullOrEmpty(type, "search type cannot be null or whitespace");
+
+        try {
+            if( type.getBytes("UTF-8").length > MAX_TYPE_SIZE) {
+                throw new IllegalArgumentException("Search type string size must be less than "+MAX_TYPE_SIZE+" UTF-8 chars");
```
MrCreosote: Line > 100 chars

lib/src/kbasesearchengine/system/SearchObjectType.java
```diff
@@ -12,20 +13,30 @@
...
+     * The first character must be a letter. The maximum size is {@link SearchObjectType#MAX_TYPE_SIZE}.
      * @param version the version of the type.
      * @throws IllegalArgumentException if the type name is null or whitespace or the version is
      * less than one.
      */
     public SearchObjectType(final String type, final int version) {
         Utils.notNullOrEmpty(type, "search type cannot be null or whitespace");
+
+        try {
+            if( type.getBytes("UTF-8").length > MAX_TYPE_SIZE) {
```
MrCreosote: Can you add a test for this new section?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -279,55 +281,118 @@ public void testGenome() throws Exception {
...
+        // null list of types
+        Set<GUID> guids = indexStorage.searchIds(null,
+                MatchFilter.create().withAccessGroupId(1),
+                null, AccessFilter.create().withAdmin(true));
+    }
+
+    @Test
+    public void testMultiTypeSearchValidation2() throws Exception {
+
+        // empty list
```
MrCreosote: Will this test pass if it runs first? It looks like it expects that at least one object is indexed, but I think all the tests index their own objects.

Really we should clear the database between every test so that tests are ensured to be independent.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -279,55 +281,118 @@ public void testGenome() throws Exception {
...
+        objectTypes.add(null);
+        objectTypes.add("Narrative");
+        Set<GUID> guids = indexStorage.searchIds(objectTypes,
+                MatchFilter.create().withAccessGroupId(1),
+                null, AccessFilter.create().withAdmin(true));
+    }
+
+    @Test
+    public void testMultiTypeSearchValidation4() throws Exception {
+        List<String> objectTypes;
```
MrCreosote: Same question as testMultiTypeSearchValidation2

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -279,55 +281,118 @@ public void testGenome() throws Exception {
...
+
+        // list containing a mix of null and non-null types
+        List<String> objectTypes = new ArrayList<>();
+        objectTypes.add(null);
+        objectTypes.add("Narrative");
+        Set<GUID> guids = indexStorage.searchIds(objectTypes,
+                MatchFilter.create().withAccessGroupId(1),
+                null, AccessFilter.create().withAdmin(true));
+    }
+
```
MrCreosote: Missing test for max number of types?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: If I correctly understood how prepare() is implemented, we have at least 2 objects being indexed before the tests are executed.

lib/src/kbasesearchengine/main/SearchMethods.java
arfathpasha: Yep. I'm thinking of writing tests that don't just test or correctness but for quality of results returned.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Added.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: I just reread prepare() and I'm not seeing any indexing that happens as part of that method. There's indexObject calls embedded in an anonymous subclass of ObjectLookupProvider that's declared in prepare() but that won't be executed as part of the prepare method(). Am I missing something?

arfathpasha: You were right and I was being lazy. I fixed the side effects that were affecting the tests. The database is now being cleared.

MrCreosote:
Looks like the retrier test needs rethinking, even a slop that's almost as big as the time difference is too little. Rerunning for now

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -79,6 +80,16 @@
     public static final int PUBLIC_ACCESS_GROUP = -1;
     public static final int ADMIN_ACCESS_GROUP = -2;
 
+    /** ElasticSearch max length of an HTTP URL defaults to 4kb.
```
MrCreosote: So this line still seems like it applies to the old field that was the url size rather than this new field that's the max # of objects. I'd expect something like 

"The maximum number of types that can be specified when restricting the search to specific types"

or something like that...

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -126,9 +137,47 @@ public void dropData() throws IOException {
...
+     * list 1 represents a list that would map to any index pattern (search unconstrained by type)
+     * list 2 represents a list that would map to one or more specific index patterns (constrained search)
+     *
+     *
+     * @param objectTypes a list of object types.
+     * @throws IOException if list is invalid.
+     */
+    private void validateObjectTypes(List<String> objectTypes) throws IOException {
+
+        if(objectTypes == null) {
```
MrCreosote: space after if, more below

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1372,8 +1421,28 @@ private FoundHits queryHits(
...
             sort.add(sortOrderWrapper);
         }
-        String indexName = objectType == null ? getAnyIndexPattern() : checkIndex(objectType);
+
+        validateObjectTypes(objectTypes);
+
+        String indexName;
+
+        // search unconstrained by object type
+        if( objectTypes.isEmpty()) {
```
MrCreosote: space after if, no space inside parens

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -267,6 +268,12 @@ public SearchTypesOutput searchTypes(SearchTypesInput params, String user) throw
...
+        // validate input
+        if(params.getObjectTypes() == null) {
```
MrCreosote: space after if

lib/src/kbasesearchengine/system/SearchObjectType.java
```diff
@@ -12,20 +13,31 @@
...
+     * The first character must be a letter. The maximum size is {@link SearchObjectType#MAX_TYPE_SIZE}.
      * @param version the version of the type.
      * @throws IllegalArgumentException if the type name is null or whitespace or the version is
      * less than one.
      */
     public SearchObjectType(final String type, final int version) {
         Utils.notNullOrEmpty(type, "search type cannot be null or whitespace");
+
+        try {
+            if(type.getBytes("UTF-8").length > MAX_TYPE_SIZE) {
```
MrCreosote: space after if

lib/src/kbasesearchengine/system/SearchObjectType.java
```diff
@@ -12,20 +13,31 @@
...
+                throw new IllegalArgumentException("Search type string size must be less than " +
+                        MAX_TYPE_SIZE + " UTF-8 chars");
+            }
+        } catch (UnsupportedEncodingException ex ) {
+            throw new IllegalArgumentException("Unsupported encoding for search type string: "+ex.getMessage());
```
MrCreosote: spaces around operators

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -282,55 +295,134 @@ public void testGenome() throws Exception {
...
+    @Test
+    public void testMultiTypeSearchValidation2() throws Exception {
+        indexObject("Genome", "genome01", new GUID("WS:1/1/1"), "MyGenome.1");
+
+        // empty list
+        Set<GUID> guids = indexStorage.searchIds(new ArrayList<String>(),
+                MatchFilter.create().withAccessGroupId(1),
+                null, AccessFilter.create().withAdmin(true));
+
+        Assert.assertTrue(guids.size() > 1);
```
MrCreosote: Should always be 1

MrCreosote: Most of the newer tests use the assertThat("message", got, is(expected)); form, I'd like to start moving towards that

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -282,55 +295,134 @@ public void testGenome() throws Exception {
         Assert.assertEquals("1", "" + assemblyIndex.getKeyProperties().get("contigs"));
         System.out.println("*** end testGenome***");
     }
+
+
+    @Test(expected = IllegalArgumentException.class)
```
MrCreosote: I realize this way is easier but we really should check that the exception message is correct

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -282,55 +295,134 @@ public void testGenome() throws Exception {
...
+        indexObject("Assembly", "assembly01", new GUID("WS:1/4/1"), "MyAssembly.1");
+
+        // saerch for Genome and Assembly types
+        Set<GUID> guids = indexStorage.searchIds(ImmutableList.of("Genome", "Assembly"),
+                MatchFilter.create().withAccessGroupId(1),
+                  null, AccessFilter.create().withAdmin(true));
+
+        Assert.assertTrue(guids.contains(GUID.fromRef("WS", "1/3/1")));
+        Assert.assertTrue(guids.contains(GUID.fromRef("WS", "1/4/1")));
+    }
```
MrCreosote: I think a test to show that if a type is not in the type list objects of that type don't show up in the search is needed.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Actually, I was right. When a genome01 object is created in a test, the lookup provider object's resolveRefs method gets called in prepare() and this populates the index with two more referenced objects

lib/src/kbasesearchengine/main/SearchMethods.java
arfathpasha: Fixed. I have intellij doing this for me now. Should not happen again.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Done.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/system/SearchObjectType.java
arfathpasha: Fixed.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Done.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: This turns out to be 3 because of the 2 additional referenced objects that get populated by the lookup provider when a genome object is indexed

arfathpasha: Done.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Test has been strengthened.

MrCreosote: Unless I'm reading this test incorrectly it seems to be showing that a non-requested type *is* in the results which is not expected

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: But that doesn't help if a test that doesn't index any objects is run first, right? It'll still fail because no objects are indexed. The lookup provider just adds more objects the first time a test indexes a genome, so it doesn't affect tests that don't index anything but expect things to be indexed.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: The side effect from the ObjectLookupProvider implementation in the prepare method is causing this. The test is correct and I have tried to clarify this doubt by breaking up the test into three stages - {Genome}, {Genome, Assembly}, {Genome, Assembly, AssemblyContig}. Keep in mind that Assemby and Assembly contig objects get indexed as a side-effect to indexing the Genome object. This test class needs to be cleaned up to remove this ugly side-effect.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: testMultiTypeSearchValidation2() indexes a genome object which causes that awful implementation in prepare to index two more objects.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Clarification from OOB conversation: The original test implementation indexed a Genome, which also causes an Assembly and an AssemblyContig to be indexed (I assume this so the the Genome reference works - if the Assembly doesn't get indexed I presume the Genome will fail to index[?]). Then the test indexes *another* Assembly and AssemblyContig. Hence when the test searches for assemblies and genomes, it gets back 3 objects vs. the intuitive two. The original comment that one of the objects is an AssemblyContig is incorrect.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Well it does now. When this comment chain started it didn't.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Now the test just indexes a Genome (causing the A and AC to be indexed) and the results make more sense.

PR #125, opened 2018-01-12
SCT-328 Add CLI minimal spec generator

MrCreosote:
The CLI argument parsing & handling portion was tested manually. Need to add tests for that at some point, but not right now.


MrCreosote:
I considered that, but this CLI will be used by maybe 5 people, and they should all be expert KBase admins, so I didn't think it was worth the trouble - e.g. YAGNI. My feeling is that if the people using the CLI ask for it we could add it but otherwise I don't think it's worth the effort.

If the CLI were intended for general use I'd absolutely agree...

PR #127, opened 2018-01-16
Use embedded auth server for tests

MrCreosote:
No longer any need for tokens in the test config when running ant test directly.
`ant test -Djars.dir=/home/crusherofheads/localgit/jars/lib/jars -Dtest.cfg=test_local/test.cfg`

Running kb-sdk test still requires a token.


MrCreosote:
next step is updating .travis.yml

MrCreosote:
I don't think there's any test documentation at all currently. We should probably do a docs push to add to the documentation @arfathpasha has written at some point.

PR #129, opened 2018-01-18
Update travis config

MrCreosote:
Might take a few tries to get this working


arfathpasha:
Only problem I see with this approach is that stuff can change with the docker env and if travis config is not updated to keep it in sync, travis tests will not reflect these changes.

PR #132, opened 2018-01-24
Add highlighting from query terms [SCT-415]

tiramisu24:
-add highlighting options for search queries


MrCreosote:
> -return highlighted fields without the key.
What does this mean?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1057,12 +1058,21 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
+                ImmutableMap.of("guid", ids.stream().map(u -> u.toString()).collect(Collectors.toList())))));
         //doc.put("_source", Arrays.asList("ojson"));
+        Map<String, Object> highlight =
+                ImmutableMap.of("fields",
+                        ImmutableMap.of("*",
+                                ImmutableMap.of("require_field_match", false)));
+        Map<String, Object> doc = new LinkedHashMap<>();
+        doc.put("query", query);
+
+        if(pp.objectHighlight){
```
MrCreosote: Space after if and before {

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1077,22 +1087,39 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
-                    pp.objectData, pp.objectDataIncludes);
+            Map<String, ArrayList> highlightRes = (Map<String, ArrayList>) hit.get("highlight");
+            ObjectData item = buildObjectData(obj, editHighlightedKeys(highlightRes), pp);
             ret.add(item);
         }
         return ret;
     }
 
+    private Map<String, ArrayList> editHighlightedKeys(Map<String, ArrayList> highlightRes){
+        if(highlightRes == null){
```
MrCreosote: spaces

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1077,22 +1087,39 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
+            return null;
+        }
+        Map<String, ArrayList> highlightResEdited = new HashMap<>();
+        for(String key : highlightRes.keySet()){
```
MrCreosote: spaces

MrCreosote: not going to make any more spaces comments but there's more in this PR

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1428,13 +1459,18 @@ private FoundHits queryHits(
...
                                            ImmutableMap.of("bool",
                                               ImmutableMap.of("should", shouldList)))));
+        Map<String, Object> highlight =
+                        ImmutableMap.of("fields",
+                                ImmutableMap.of("*",
+                                        ImmutableMap.of("require_field_match", false)));
 
         Map<String, Object> doc = new LinkedHashMap<>();
         doc.put("query", query);
+        doc.put("highlight", highlight);
```
MrCreosote: Why is this not selectable like in search by ids?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -22,6 +22,7 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.apache.commons.io.IOUtils;
```
MrCreosote: Looks like this is unused

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1057,12 +1058,21 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
 
-        Map<String, Object> doc = ImmutableMap.of("query",
-                                     ImmutableMap.of("bool",
+        Map<String, Object> query = ImmutableMap.of("bool",
                                         ImmutableMap.of("filter",
                                            ImmutableMap.of("terms",
-                ImmutableMap.of("guid", ids.stream().map(u -> u.toString()).collect(Collectors.toList()))))));
+                ImmutableMap.of("guid", ids.stream().map(u -> u.toString()).collect(Collectors.toList())))));
         //doc.put("_source", Arrays.asList("ojson"));
+        Map<String, Object> highlight =
```
MrCreosote: This appears to be duplicated in a couple places - DRY it up with a method

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1077,22 +1087,39 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
     }
 
+    private Map<String, ArrayList> editHighlightedKeys(Map<String, ArrayList> highlightRes){
+        if(highlightRes == null){
+            return null;
+        }
+        Map<String, ArrayList> highlightResEdited = new HashMap<>();
+        for(String key : highlightRes.keySet()){
+            String newKey = key;
+            if(key.startsWith("key.")){
```
MrCreosote: Is there functionality like this anywhere else in the code? I seem to remember something similar and if so we should consolidate into a method.

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -526,5 +548,16 @@ public Builder withKeyProperty(final String key, final String property) {
             keyProps.put(key, property);
             return this;
         }
+
+        /** Adds the highlight fields to the object.
+         * @param highlight the map of fields returned from elasticsearch.
+         * @return this builder.
+         */
+        public Builder withNullableHighlight(final Map<String, ArrayList> highlight) {
```
MrCreosote: Please follow the example of the other builders and don't add maps directly. Add key by key and enforce reasonable constraints on the key and value. In particular:
1) No nulls or empty strings
2) No unparameterized generics
3) Don't specify the List implementation

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -179,6 +185,13 @@ public GUID getGUID() {
         return keyProps;
     }
 
+    /** Get the highlighted fields and snippet of the object generated by elasticsearch, if present.
+     * @return the highlighted fields with corresponding snipped to be highlighted.
+     */
+    public Map<String, ArrayList> getHighlight() { return highlight; }
```
MrCreosote: Don't leak implementation details outside of the API.

Always have the method body on a separate line.

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: Needs tests in ObjectDataTest

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1077,22 +1087,39 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
+                //truncate key. from key
+                newKey = key.substring(4);
+            }
+            highlightResEdited.put(newKey, highlightRes.get(key));
```
MrCreosote: What are the typical contents of a highlight map?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: okay. actually since we are talking about duplicates. Why does fn searchIds even exits. Its the same as searchObjects just with null passed for PostProcessing. The naming of the methods are also super weird. getObjectsByIds is just search but searchIds doesnt even search on ids.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: is there an auto space detection in eclipse or something? how are you able to spot these spaces/lack of spaces??

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: I dont think so? there is functionality that that add the "key." to fields but not one to strip the key

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: No idea, before my time.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: They stick out like a sore thumb to me

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Looks like they're only used in tests

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: There's something vaguely similar in buildObjectData

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: so you mean separate it out into key and value? Can I directly add the ArrayList?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: highlight: [{"field" : "<em>matched term</em>"},{"field2:" : "<em>matched term</em>"}] for each object.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: weird I thought I had also put that as selectable. Will be refactoring that and into a method

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: I should only need to modify the existing tests to include the field right?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: is matched term a list? that's what it looks like in ObjectData

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: I personally would probably do something like

```
public Builder withHighLight(final String field, final String highlight) {
    Utils.notNullOrEmpty(field, ...
    Utils.notNullOrEmpty(highlight, ...
    if (!higlights.containsKey(field)) {
        highlights.put(field, new HashSet<>());
    }
    highlights.get(field).add(highlight);
    return this;
}
```
I suppose you could do 
public Builder withHighLight(final String field, final List<String> highlights) {
    Utils.notNullOrEmpty(field, ...
    Utils.notNull(highlights, ...
    Utils.noNulls(highlights, ...
     this.highlights.put(field, highlights)
    return this;
 }

MrCreosote: yep, take a look at the current tests for how to test the additional field correctly

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: the whole list is the list. The object may have multiple matches and not all in the same field

MrCreosote: If highlight in your comment above is the map key, why are you removing `key.` from the front? You really need to go through the list for each highlight key, go through the each map in the list, and edit those, since those are where the search keys are

tiramisu24: Sorry I just realized you were asking about the field stored in ObjectData. 
That goes: 
{"field1":["match1], "field2":["match1", "match2"]}
so the keys are the fields and matches show up in arrays.

MrCreosote: But it's the same structure from elastic, right? Currently the code just adds it to the ObjectData after stripping of `key.`:

```
Map<String, ArrayList> highlightRes = (Map<String, ArrayList>) hit.get("highlight");
ObjectData item = buildObjectData(obj, editHighlightedKeys(highlightRes), pp);
```

tiramisu24: Yes its the same as elastic. First pattern was wrong

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: Actually the latter implementation allows empty strings in the list

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: So I looked at all the queries and I could pull out the 2-3 lines that are similar but I feel like it reads less well. The only things staying the same is bool, then filter. What do you think?

MrCreosote: Hmm, our opinions differ on this one I guess. I think it's easier to read as the highlight query creation is abstracted away. More importantly, if the highlight query needs to change we can do it in one place vs 2+.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: I cant guarantee I can catch all of them. Feels like Looking for air bubbles in a frozen lake.

MrCreosote: NOT ACCEPTABLE

tiramisu24: or maybe if there are enough thumbs sticking out it will look flat

MrCreosote:
>> -return highlighted fields without the key.

> What does this mean?

Still wondering about this

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -7,18 +7,7 @@
...
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
-import java.util.LinkedList;
+import java.util.*;
```
MrCreosote: Please don't use wildcard imports

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1052,18 +1041,32 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
         pp.objectKeys = true;
         return getObjectsByIds(ids, pp);
     }
-    
+
+    private Map<String, Object> createHighlightQuery(){
+        Map<String, Object> highlight =
```
MrCreosote: Trivial, but you could just return directly rather than creating a variable.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1052,18 +1041,32 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
                                         ImmutableMap.of("filter",
                                            ImmutableMap.of("terms",
-                ImmutableMap.of("guid", ids.stream().map(u -> u.toString()).collect(Collectors.toList()))))));
+                ImmutableMap.of("guid", ids.stream().map(u -> u.toString()).collect(Collectors.toList())))));
         //doc.put("_source", Arrays.asList("ojson"));
 
+        Map<String, Object> doc = new LinkedHashMap<>();
+        doc.put("query", query);
+
+        if(Objects.nonNull(pp) && pp.objectHighlight) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1077,22 +1080,20 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
     }
 
     private ObjectData buildObjectData(
             final Map<String, Object> obj,
-            final boolean info,
-            final boolean keys, 
-            final boolean json,
-            final List<String> objectDataIncludes) {
+            final Map<String, ArrayList> highlight,
+            PostProcessing pp) {
```
MrCreosote: make final like the other args

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1132,12 +1133,27 @@ private ObjectData buildObjectData(
                     } else {
                         textValue = String.valueOf(objValue);
                     }
-                    b.withKeyProperty(key.substring(4), textValue);
+                    b.withKeyProperty(stripKeyPrefix(key), textValue);
                 }
             }
         }
+        if(pp.objectHighlight) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1132,12 +1133,27 @@ private ObjectData buildObjectData(
...
+                    newKey = stripKeyPrefix(key);
+                }
+                b.withHighlight(newKey, highlight.get(key));
+            }
+        }
+
         return b.build();
     }
+
+    private String stripKeyPrefix(String key){
```
MrCreosote: final arg

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1132,12 +1133,27 @@ private ObjectData buildObjectData(
                     } else {
                         textValue = String.valueOf(objValue);
                     }
-                    b.withKeyProperty(key.substring(4), textValue);
+                    b.withKeyProperty(stripKeyPrefix(key), textValue);
                 }
             }
         }
+        if(pp.objectHighlight) {
+            for(String key : highlight.keySet()) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1132,12 +1133,27 @@ private ObjectData buildObjectData(
...
+                String newKey = key;
+                if(key.startsWith("key.")) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1431,10 +1447,13 @@ private FoundHits queryHits(
 
         Map<String, Object> doc = new LinkedHashMap<>();
         doc.put("query", query);
+        if(Objects.nonNull(pp) && pp.objectHighlight) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1475,30 +1494,32 @@ private FoundHits queryHits(
 
         Response resp = makeRequest("GET", urlPath, ImmutableMap.copyOf(doc));
         @SuppressWarnings("unchecked")
+//        String results = IOUtils.toString(resp.getEntity().getContent(), "UTF-8");
```
MrCreosote: delete commented out code

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1475,30 +1494,32 @@ private FoundHits queryHits(
...
         if (loadObjects) {
-            ret.objects = new ArrayList<ObjectData>();
+            ret.objects = new ArrayList<>();
         }
         @SuppressWarnings("unchecked")
         List<Map<String, Object>> hitList = (List<Map<String, Object>>) hitMap.get("hits");
         for (Map<String, Object> hit : hitList) {
             @SuppressWarnings("unchecked")
             Map<String, Object> obj = (Map<String, Object>) hit.get("_source");
+            Map<String, ArrayList> highlightRes = (Map<String, ArrayList>) hit.get("highlight");
```
MrCreosote: Don't specify the implementation

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -1,6 +1,8 @@
 package kbasesearchengine.search;
 
+import java.lang.reflect.Array;
```
MrCreosote: This looks unused

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -33,7 +35,9 @@
     private final Optional<Object> parentData;
     private final Optional<Object> data;
     private final Map<String, String> keyProps;
-    
+    private final Map<String, ArrayList> highlight;
```
MrCreosote: Don't specify the implementation - this applies in several places in this class.

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -526,5 +548,18 @@ public Builder withKeyProperty(final String key, final String property) {
...
         }
+
+        /** Adds the highlight fields to the object.
+         * @param highlight the map of fields returned from elasticsearch.
+         * @return this builder.
+         */
+        public Builder withHighlight(final String field, final ArrayList highlight) {
+            Utils.notNullOrEmpty(field, "field cannot be null or whitespace");
+            Utils.noNulls(highlight, "highlight cannot be null");
+            if(highlight.size() > 0) {
```
MrCreosote: space

lib/src/kbasesearchengine/search/ObjectData.java
```diff
@@ -526,5 +548,18 @@ public Builder withKeyProperty(final String key, final String property) {
...
             return this;
         }
+
+        /** Adds the highlight fields to the object.
+         * @param highlight the map of fields returned from elasticsearch.
+         * @return this builder.
+         */
+        public Builder withHighlight(final String field, final ArrayList highlight) {
+            Utils.notNullOrEmpty(field, "field cannot be null or whitespace");
+            Utils.noNulls(highlight, "highlight cannot be null");
```
MrCreosote: Check for any whitespace only highlight entries and throw an exception if so

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: No non-parameterized Generics - this applies in several places in this class.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: No non-parameterized Generics

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Make variable final.

This applies all over in this PR - always make variables final unless there's a good reason for them not to be.

test/src/kbasesearchengine/test/search/ObjectDataTest.java
```diff
@@ -8,6 +8,8 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.ArrayList;
+import java.util.Arrays;
```
MrCreosote: This doesn't test any failure conditions

test/src/kbasesearchengine/test/search/ObjectDataTest.java
```diff
@@ -150,6 +158,10 @@ public void buildMaximal() {
         keyprops.put("null", null);
         keyprops.put("ws", "   \t   \n   ");
         assertThat("incorrect key props", od.getKeyProperties(), is(keyprops));
+        final Map<String, ArrayList> highlight = new HashMap<>();
+        highlight.put("field", new ArrayList<>(Arrays.asList("match")));
+        assertThat("incorrect highlight field", od.getHighlight().keySet(), is(highlight.keySet()));
```
MrCreosote: Just test the highlight output directly

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: I don't see a test for this path

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1000,4 +1004,45 @@ public void testLookupInKey() throws Exception{
...
+
+    @Test
+    public void addHighlighting() throws Exception {
+        GUID guid1 = new GUID("WS:11/1/2");
+        GUID guid2 = new GUID("WS:11/2/2");
+        GUID guid3 = new GUID("WS:11/3/2");
+        try {
+            prepareTestMultiwordSearch(guid1, guid2, guid3);
+        } catch(Exception e) {
+            //TODO: remove items from each test after completion. Will error if run all sequence in order due to version conflict
```
MrCreosote: The elastic indexes are cleared after every test

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: okay

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: after each test in the the test file? I thought it runs the prep functions and then just runs the tests in sequence

test/src/kbasesearchengine/test/search/ObjectDataTest.java
tiramisu24: what do you mean?

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: Do you mean I shouldn't do something like 
private Map<String, ArrayList> highlight = new HashMap<>();

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: could you tell me how is this a non-parameterized Generic? B/c of the Map?

MrCreosote: ArrayList is a generic class, and it's not parameterized.

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: No, it means you sholdn't use generic classes without parameterizing them.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: Yep, see the method marked with the @After annotation

test/src/kbasesearchengine/test/search/ObjectDataTest.java
MrCreosote: assertThat("message", od.getHighlight(), is(expected));

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: so I should do something like this? 
private Map<String, ArrayList<String>> highlight = new HashMap<>();

MrCreosote: No, ArrayList is a generic class, and it's not parameterized.

tiramisu24: wtf wrong comment posted. I meant

private Map<String, ArrayList<String>> highlight = new HashMap<>();

tiramisu24: interesting so in github you need to escape <> characters otherwise the inside doesnt show

MrCreosote: yes. also as I mentioned elsewhere don't specify the list implementation

tiramisu24: what do you mean by not specifying list implementation? Do List<String> instead of ArrayList<String>?

MrCreosote: Right, unless you need a property of a specific implementation, which you usually don't

tiramisu24: okay it makes sense now

test/src/kbasesearchengine/test/search/ObjectDataTest.java
tiramisu24: what failure conditions?

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: buildMaximal in ObjectDataTest tests it?

lib/src/kbasesearchengine/search/ObjectData.java
tiramisu24: is there a need to iterate through the highlights and check? b/c if elastic search highlight is returning just whitespaces its broken

MrCreosote: Yes, there's no guarantee that this class will always be paired with elasticsearch. A class should not consider implementation details of its dependencies in its design, because those implementation details may change.

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: It doesn't test the case that highlight.size() < 1 as far as I can tell

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: Also elasticsearch isn't really a dependency anyway - this class doesn't know elasticsearch exists.

test/src/kbasesearchengine/test/search/ObjectDataTest.java
MrCreosote: The null and whitespace checks

tiramisu24:
-return highlighted fields without the key.

What does this mean?

Still wondering about this

return all fields without any of them being called key.something.

lib/src/kbasesearchengine/search/ObjectData.java
MrCreosote: The other problem I just realized about this implementation is that because the list is passed in and returned as is, it allows the internal state to be modified from outside the class after creation, breaking the classes' immutability

tiramisu24: does this mean that I should return an unmodifiableMap in fn getHighlight?

MrCreosote: I would just build the entire structure so it's immutable in the constructor. Then you can return whatever and be safe.

The other approach I've used is instead of returning the map, have getKeys() and getValue(key) methods (appropriately named) so the value class doesn't have to expose the map.

tiramisu24: okay. its an unmodifiableMap in the constructor. 

As far as I understand, the current MVP wants me to chuck the whole map out so the second approach won't make sense.

MrCreosote: You could still do it that way, you'd just rebuild the map from the class.

tiramisu24: I could but isnt that what the unmodifiable map already doing in addition to not allowing you to change the outputted map? If my only use case is returning everything, then it seemed kind of unnecessary to do the rebuild myself.

MrCreosote: I said could, not should :P

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1057,17 +1058,28 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
+    }
+
     @Override
-    public List<ObjectData> getObjectsByIds(Set<GUID> ids, PostProcessing pp) 
+    public List<ObjectData> getObjectsByIds(Set<GUID> ids, final PostProcessing pp)
             throws IOException {
 
-        Map<String, Object> doc = ImmutableMap.of("query",
-                                     ImmutableMap.of("bool",
+        Map<String, Object> query = ImmutableMap.of("bool",
```
MrCreosote: Make vars final unless there's a good reason not to

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1111,9 +1121,9 @@ private ObjectData buildObjectData(
                     (Integer) obj.get(SEARCH_OBJ_TYPE_VER)));
             // sometimes this is a long, sometimes it's an int
             b.withNullableTimestamp(Instant.ofEpochMilli(
-                    ((Number) obj.get(OBJ_TIMESTAMP)).longValue()));
+            ((Number) obj.get(OBJ_TIMESTAMP)).longValue()));
```
MrCreosote: Indent

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1086,4 +1090,48 @@ public void testLookupInKey() throws Exception{
...
+        filter.withNullableFullTextInAll("multiWordInSearchMethod1 multiWordInSearchMethod2");
+
+        FoundHits hits = indexStorage.searchObjects(empty, filter.build(),sorting, accessFilter, null, pp);
+        Map<String, List<String>> hitRes = hits.objects.get(0).getHighlight();
+
+        Map<String, List<String>> result1 = new HashMap<>();
+        result1.put("prop1", Arrays.asList("<em>multiWordInSearchMethod1</em> <em>multiWordInSearchMethod2</em>"));
+        assertThat("Incorrect highlighting", hitRes, is(result1) );
+
+        //searchIds is a wrapper around queryHits and does not r<em>WS:11/1/2</em>eturn object data and so will not be highlighted
```
MrCreosote: What's with the embedded guid...?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1057,17 +1058,28 @@ public void unpublishObjectsExternally(Set<GUID> guids, int accessGroupId) throw
...
+
+    private Map<String, Object> createHighlightQuery(){
+        return  ImmutableMap.of("fields",
+                    ImmutableMap.of("*",
+                            ImmutableMap.of("require_field_match", false)));
+    }
+
     @Override
-    public List<ObjectData> getObjectsByIds(Set<GUID> ids, PostProcessing pp) 
+    public List<ObjectData> getObjectsByIds(Set<GUID> ids, final PostProcessing pp)
```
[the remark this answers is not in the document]
tiramisu24: dang it I thought I caught all of them already

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: accidentally copied and pasted that in there

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1086,4 +1090,48 @@ public void testLookupInKey() throws Exception{
...
+
+    @Test
+    public void addHighlighting() throws Exception {
+        GUID guid1 = new GUID("WS:11/1/2");
+        GUID guid2 = new GUID("WS:11/2/2");
+        GUID guid3 = new GUID("WS:11/3/2");
+//        try {
+            prepareTestMultiwordSearch(guid1, guid2, guid3);
+//        } catch(Exception e) {
+//            //TODO: remove items from each test after completion. Will error if run all sequence in order due to version conflict
```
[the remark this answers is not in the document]
tiramisu24: yep when did that change? I remember it didnt use to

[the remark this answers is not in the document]
MrCreosote: @arfathpasha fixed that a couple weeks or so ago

PR #133, opened 2018-01-26
SCT-427 Add builder for StoredStatusEvent with worker codes

MrCreosote:
In preparation for adding the ability to assign events to a subset of workers.

Worker codes are currently unused elsewhere.

### DIFF
--- a/lib/src/kbasesearchengine/events/StoredStatusEvent.java
+++ b/lib/src/kbasesearchengine/events/StoredStatusEvent.java
@@ -21,7 +21,7 @@ public class StoredStatusEvent implements StatusEventWithId {
     private final StatusEventProcessingState state;
     private final Optional<Instant> updateTime;
     private final Optional<String> updater;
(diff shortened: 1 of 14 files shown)


lib/src/kbasesearchengine/events/StoredStatusEvent.java
```diff
@@ -18,36 +21,21 @@
     private final StatusEventProcessingState state;
     private final Optional<Instant> updateTime;
     private final Optional<String> updater;
+    private final Set<String> workerCodes;
```
arfathpasha: It will be helpful if you can document what workerCodes are and what they are being used for. 

Same for updater. The other attribs are quite unambiguous.

MrCreosote: Is the javadoc for the getters and builder with() methods not clear? I just reread them and they seem ok to me, but I'm happy to incorporate improvements in a future pr

PR #134, opened 2018-01-27
Fix class path

MrCreosote:
Apparently some instances of eclipse can't handle the absolute path.


MrCreosote:
@tiramisu24

tiramisu24:
@MrCreosote that is how I have been changing my .classpath

PR #135, opened 2018-01-29
SCT-398 - Add CLI command to rerun failed events

arfathpasha:
- I have used the "unshare" method in indexer worker to delete indexes for objects that are found to be deleted while rerunning failed events. 

- Minor detail - while updating an event, I retain the original event timestamp. Updating the timestamp made many existing tests fail. 

- unit tests for new code to follow.


MrCreosote:
Conflicts, needs a merge

MrCreosote:
Are you planning to put the tests in a new PR, or add to this one?

arfathpasha:
I can add to this one once you OK the implementation.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        }
+
+
+        // wsid and objid of the object that the specified StatusEvent is an event for
+        long wsid = Long.valueOf(ev.getAccessGroupId().get().toString()).longValue();
+        long objid = Long.valueOf(ev.getAccessGroupObjectId().get().toString()).longValue();
+
+        Map<String, Object> command;
+
+        // check if object is deleted and update state of latestEventType
```
MrCreosote: Why not just make the get object info call below, and if it throws a JsonClientException containing a string indicating the workspace or object was deleted, mutate the event to DELETE_ALL_VERSIONS?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        if( ev.isPublic().isPresent()) {
+            latestIsPublic = ev.isPublic().get().booleanValue();
+        }
+        if( ev.getNewName().isPresent()) {
+            latestName = ev.getNewName().get().toString();
+        }
+
+
+        // wsid and objid of the object that the specified StatusEvent is an event for
+        long wsid = Long.valueOf(ev.getAccessGroupId().get().toString()).longValue();
```
MrCreosote: So officially not all events have access group ids seeing as it's optional. All workspace generated events do have AGIDs. I'm wondering if the optional AGIDs should change...

If it doesn't this code needs to handle that though.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
         return String.join(";", refpath);
     }
 
+    @Override
+    public StatusEvent updateEvent(
```
MrCreosote: I think the arg can be on the same line right, since it's < 100 chars.

Arg can be final as well

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        } catch (IOException e) {
+            throw new RetriableIndexingException(e.getMessage());
+        } catch (JsonClientException e) {
+            throw new RetriableIndexingException(e.getMessage());
+        }
+
+
+        // check if name has changed and update state of lastestName
+        ObjectSpecification os = new ObjectSpecification().withRef(
+                ev.getAccessGroupId().get().toString() + "/" +
```
MrCreosote: Event may not have AGID or ObjID

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+            }
+        } catch (IOException e) {
+            throw new RetriableIndexingException(e.getMessage());
+        } catch (JsonClientException e) {
+            throw new RetriableIndexingException(e.getMessage());
+        }
+
+
+        // check if name has changed and update state of lastestName
+        ObjectSpecification os = new ObjectSpecification().withRef(
```
MrCreosote: final

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        command = new HashMap<>();
+        command.put("command", "getObjectInfo");
+        command.put("params", new GetObjectInfo3Params().withObjects(getInfoInput));
+
+        final GetObjectInfo3Results objInfo;
+        try {
+            objInfo = ws.getClient().administer(new UObject(command))
+                    .asClassInstance(GetObjectInfo3Results.class);
+            String name = objInfo.getInfos().get(0).getE2();
+            if ( !Utils.isNullOrEmpty(name) ) {
```
MrCreosote: I don't think there's any case where this can be false, is there?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+            command.put("command", "getWorkspaceInfo");
+            command.put("params", new WorkspaceIdentity()
+                    .withId((long) ev.getAccessGroupId().get()));
+
+            final String isPublic;
+
+            isPublic = ws.getClient().administer(new UObject(command))
+                    .asClassInstance(WS_INFO_TYPEREF).getE6();
+
+            if ( !Utils.isNullOrEmpty(isPublic) ) {
```
MrCreosote: Virtually sure this can never be false

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        }
+
+        // check if permissions have changed and update state of isPublic
+        try {
+            command = new HashMap<>();
+            command.put("command", "getWorkspaceInfo");
+            command.put("params", new WorkspaceIdentity()
+                    .withId((long) ev.getAccessGroupId().get()));
+
+            final String isPublic;
```
MrCreosote: Why the separate declaration & initialization?

lib/src/kbasesearchengine/events/storage/MongoDBStatusEventStorage.java
```diff
@@ -317,4 +317,24 @@ public boolean setProcessingState(
         return Optional.of(toEvent(ret));
     }
 
+    @Override
```
MrCreosote: Here I think some discussion is warranted. From an implementation standpoint, IMO this should just be done with a mongo query that finds events with FAIL and sets them to UNPROC.

The bigger issue is whether resetting all events makes sense. There's probably some events that are never going to pass and that should not be rerun. I haven't completely thought this through yet, but there may be events that are dangerous to rerun. Thoughts?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -55,6 +56,9 @@
 import kbasesearchengine.system.TypeStorage;
 import kbasesearchengine.tools.Utils;
 import org.apache.commons.io.FileUtils;
+import workspace.GetObjectInfo3Params;
```
MrCreosote: I don't think any of these are used?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -403,35 +407,40 @@ private EventHandler getEventHandler(final String storageCode)
...
-                            ev.isPublic().get(), null, new LinkedList<>());
+                    indexObject(pguid, updatedEvent.getStorageObjectType().get(), updatedEvent.getTimestamp(),
+                            updatedEvent.isPublic().get(), null, new LinkedList<>());
                 }
                 break;
             // currently unused
-//            case DELETED:
-//                unshare(ev.toGUID(), ev.getAccessGroupId().get());
-//                break;
+            case DELETED:
```
MrCreosote: Why not DELETE_ALL_VERSIONS? Unshare has to do with data palettes, not deletion (and I'll grant you, the name and the whole data palettes situation is really confusing)

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -403,35 +407,40 @@ private EventHandler getEventHandler(final String storageCode)
 
     public void processOneEvent(final StatusEvent ev)
             throws IndexingException, InterruptedException, RetriableIndexingException {
+
+        // update event to reflect the latest state of object for which the specified StatusEvent is an event for
+        EventHandler handler = getEventHandler(ev.getStorageCode());
```
MrCreosote: I think that any exceptions thrown here should be handled correctly, but can you confirm?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: The reason for the optional AGIDs was that we may add data sources to the index in the future that don't have access groups, IIRC.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+            if ( !Utils.isNullOrEmpty(name) ) {
+                latestName = name;
+            }
+        } catch (IOException e) {
+            throw handleException(e);
+        } catch (JsonClientException e) {
+            throw handleException(e);
+        }
+
+        // check if permissions have changed and update state of isPublic
```
MrCreosote: Should add this to the returned object data at some point...

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+                latestIsPublic = (isPublic == "n") ? true: false;
+
+            }
+        } catch (IOException | JsonClientException ex) {
+            throw new RetriableIndexingException(ex.getMessage());
+        }
+
+        StatusEvent updatedEvent = StatusEvent.
+                getBuilder(ev.getStorageObjectType().get(), ev.getTimestamp(), latestEventType).
+                withNullableAccessGroupID(ev.getAccessGroupId().get()).
```
MrCreosote: May not have accgrp id, object id, or version - use .orNull()

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+            if ( !Utils.isNullOrEmpty(isPublic) ) {
+                latestIsPublic = (isPublic == "n") ? true: false;
+
+            }
+        } catch (IOException | JsonClientException ex) {
+            throw new RetriableIndexingException(ex.getMessage());
+        }
+
+        StatusEvent updatedEvent = StatusEvent.
+                getBuilder(ev.getStorageObjectType().get(), ev.getTimestamp(), latestEventType).
```
MrCreosote: may not always have a storage object type, might just have a storage code

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -54,6 +55,8 @@
 import workspace.WorkspaceClient;
 import workspace.WorkspaceIdentity;
 
+import javax.rmi.CORBA.Util;
```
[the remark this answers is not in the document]
arfathpasha: Removed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+            StatusEvent ev)
+            throws IndexingException,
+                   RetriableIndexingException {
+
+        // latest states to update
+        StatusEventType latestEventType = ev.getEventType();
+        Boolean latestIsPublic = null;
+        String latestName = null;
+
+        if( ev.isPublic().isPresent()) {
```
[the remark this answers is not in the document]
arfathpasha: Fixed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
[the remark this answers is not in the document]
arfathpasha: Fixed for one. I didn't realize one id is a string and the other is a long.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: I thought about this approach but felt too hacky to depend on the an exception message substring to make a decision.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+
+
+        // wsid and objid of the object that the specified StatusEvent is an event for
+        long wsid = Long.valueOf(ev.getAccessGroupId().get().toString()).longValue();
+        long objid = Long.valueOf(ev.getAccessGroupObjectId().get().toString()).longValue();
+
+        Map<String, Object> command;
+
+        // check if object is deleted and update state of latestEventType
+        command = new HashMap<>();
```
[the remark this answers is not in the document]
arfathpasha: I have added code for this case.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        command.put("command", "listObjects");
+        command.put("params", new ListObjectsParams().withShowOnlyDeleted(wsid).withIds(Arrays.asList(wsid)));
+        final List<Tuple11> deletedObjList;
```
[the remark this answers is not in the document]
arfathpasha: Fixed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+        try {
+            deletedObjList = ws.getClient().administer(new UObject(command))
+                    .asClassInstance(List.class);
+
+            for( Tuple11 obj: deletedObjList) {
```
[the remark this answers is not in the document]
arfathpasha: Fixed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+                long deletedObjid = Long.valueOf(obj.getE1().toString()).longValue();
+                if(deletedObjid  == objid) {
+                    latestEventType = StatusEventType.DELETED;
```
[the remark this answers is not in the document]
arfathpasha: I have switched to DELETE_ALL_VERSIONS. I wasn't sure if a workspace delete operation deleted all versions or only the version it was being called on (in the case where version is specified). I have not come across much documentation for workspace delete operation.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -570,4 +573,108 @@ private static String toWSRefPath(final List<GUID> objectRefPath) {
...
+                    break;
+                }
+            }
+        } catch (IOException e) {
+            throw new RetriableIndexingException(e.getMessage());
```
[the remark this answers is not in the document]
arfathpasha: I'm don't see a HandleException in Search.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: fixed

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: This was a paranoid check since I couldn't find anything in documentation that stated either way. I have removed the check.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: Not sure I follow. The isPublic info that is being extracted from WS is put into the updated status event that is returned.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: Yet another paranoid check which I have removed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Yes it is being handled correctly.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: If the event reached the WorkspaceEventHandler, it must have the storage object type right?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
arfathpasha: Same as above. This info is being extracted in the WorkspaceEventHandler with the assumption that the event originated from the Workspace.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: Personally I'd like to update the sdk to add error codes to exceptions like the auth service. That being said, there will be times where we need to discriminate between exceptions and right now the only way to do that is via the string. We already do it in this class, so I'm ok with it. My personal opinion is that the simpler code is worth the hackiness but I leave the call up to you.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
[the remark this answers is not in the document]
MrCreosote: It does all versions: https://ci.kbase.us/services/ws/docs/Workspace.html#funcdefWorkspace.delete_objects

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: Sorry, I meant workspace side. The object_info tuple in the workspace API should tell you whether the object is public or not.

It'll wind up being a Tuple642<> though...

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: No, for instance a DELETE_ALL_VERSIONS won't. I think the majority of the events won't.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: OK, so it'll always have an accgrp id in the case of the workspace, but the other two may not be present.

lib/src/kbasesearchengine/events/storage/MongoDBStatusEventStorage.java
arfathpasha: I am updating the method to use a mongo query. Given that we are now seeking the latest state of the storage objects, I don't see how anything can go wrong with just this function of correctly indexing objects with out-of-order events. There may be other unrelated issues that constantly prevent the event from being updated or the object from being indexed. For this, we could have another CLI to delete failed events that are older than a certain threshold timestamp. I would imagine that this would be done after logs have been reviewed, an attempt to fix problems has been made and a decision to purge these events is the option of last resort.

MrCreosote: Fair enough. If there are events that turn out to be dangerous we can deal with them as they come up.

Rather than delete the events we could just  move them to a permanently failed state.

I'm expecting that we'll need some sort of UI to change states on selected events and allow more state transitions (I'm particularly thinking of PROC->UNPROC) but we can do that with mongo surgery for now.

MrCreosote: Been thinking about this a bit more, and realized there are some events that in some cases you don't want to run again because some of the state is reflected in the event type itself, i.e.

PUBLISH_ACCESS_GROUP and UNPUBLISH_ACCESS_GROUP
Since the event type determines what state is set on the records in ES, running a failed PAG after a UAG would put the records in an inconsistent state with the WS.

DELETE_ALL_VERSIONS and UNDELETE_ALL_VERSIONS
Same logic as above

DELETE_ACCESS_GROUP and UNDELETE_ACCESS_GROUP
yadda yadda

The delete events aren't a huge problem because
1) the WS doesn't emit UNDELETE_ACCESS_GROUP events
2) The UI doesn't support undelete operations

I think the other events are safe to rerun.

Obviously this again argues for removing state from the events, or at least collapsing the events in some way such that the state can be easily updated like with RENAME_OBJECT, for example.

arfathpasha: How about, 
- If the WS object is deleted and index contains the object, delete object in index
- else, delete object in index and treat failed event as a new version event

MrCreosote: Do you mean delete as in remove from the index? If so I don't think there's a need to do that; marking it deleted is fine.

IMO we shouldn't worry about deletion for now since undelete events aren't implemented by the UI. We should, however, note all the problems with deletion and undeletion (e.g. undelete needs to update the state of the index from the ws since object state may have changed since the last index, and out of order deletions/undeletions) in implementation_notes.md

There does need to be a solution for the publish events though. Maybe it's worth it to collapse those into one event so that the state can be updated.

arfathpasha:
Will be creating a new PR from develop branch

PR #141, opened 2018-01-31
Index subobjects first

MrCreosote:
If we switch to checking data object guids in the elasticsearch index
rather than access document guids to ensure an object is indexed,
threre's a race condition that the data object guid may exist while the
subobjects are still indexing. This ensures the subobjects go first and
so if the guid exists, indexing is complete.

That would mean that for every type with subobject specs there needs to be a standard spec as well.

### DIFF
--- a/test/src/kbasesearchengine/test/system/ParsingRulesSubtypeFirstComparatorTest.java
+++ b/test/src/kbasesearchengine/test/system/ParsingRulesSubtypeFirstComparatorTest.java
@@ -1,0 +1,70 @@
+package kbasesearchengine.test.system;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.Test;
+
+import kbasesearchengine.common.ObjectJsonPath;
+import kbasesearchengine.system.ObjectTypeParsingRules;
+import kbasesearchengine.system.ParsingRulesSubtypeFirstComparator;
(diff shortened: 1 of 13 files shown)


MrCreosote:
Arg forgot to include build.xml changes

MrCreosote:
Hmm, diff coverage should be 100% - the comparator is fully covered

MrCreosote:
Maybe codecov didn't process the update?

test/src/kbasesearchengine/test/system/ParsingRulesSubtypeFirstComparatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+        
+    }
+    
+    @Test
+    public void sortWithList() throws Exception {
+        final List<ObjectTypeParsingRules> rules = new ArrayList<>(
+                Arrays.asList(R1, R2, R3, R4));
+        Collections.sort(rules, new ParsingRulesSubtypeFirstComparator());
+        
+        assertThat("incorrect sort", rules, is(Arrays.asList(R2, R4, R1, R3)));
```
arfathpasha: Depending on the sort implementation the sorted order could vary.  Wondering if this test is really required since the comparator has been tested already and the validity of this test solely depends on the validity of the comparator.

MrCreosote: In this case Collections.sort() is stable, so the order should always be the same: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#sort-java.util.List-

You're right in that the test isn't **really** necessary, but since the purpose of this class in the context of the search application is to make sure the subobject rules are first in the list I threw it in.

PR #142, opened 2018-01-31
SCT-285: SearchObjectInput compile error fix

arfathpasha:


MrCreosote:
Is this it? When I recompiled there were a lot more changes

arfathpasha:
@MrCreosote I have merged this tiny change since it was definitely needed and am hoping it is all that is needed to unblock Erik and Diane. I am not aware of any other compile error besides this.

PR #145, opened 2018-02-02
SCT-508  Alter queue so verison level events block each other 

MrCreosote:
Currently version level events don't block each other in the queue, and
unlimited version level events can run at the same time per object. It
turns out that the ElasticIndexingStorage class runs a script that
touches all versions of an object when a NEW_VERSION event is processed,
and this is causing 409 conflict errors when processing two versions of
the same object at the same time.

Alter the queue so that only one NEW_VERSION event can run per object at
once.


MrCreosote:
Ok, deleting code resulting in decreased coverage is really sad

MrCreosote:
Do you want to go over it in a call?

arfathpasha:
Yes, let's do that. We need to talk about 398 too as I am blocked. I am available all day.

PR #147, opened 2018-02-04
SCT-515: Add media and model types

JamesJeffryes:
Also converts other indexer types to yaml and adds tests for new types


build.xml
```diff
@@ -173,6 +173,7 @@ java -cp ${jar.absolute.path}:${lib.classpath} kbasesearchengine.tools.SearchToo
             <include name="kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java"/>
             <include name="kbasesearchengine/test/integration/IndexerIntegrationTest.java"/>
             <include name="kbasesearchengine/test/integration/IndexerWorkerIntegrationTest.java"/>
+            <include name="kbasesearchengine/test/parse/ObjectParserTest.java"/>
```
arfathpasha: Minor - the first line of the fileset has kbasesearchengine/test/parse/**Test.java. We should either remove this line or remove the first line and add all parse package tests individually to keep with the theme.

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,518 @@
...
+        final String jsonResource = "genome01";
+        final String type = "GenomeFeature";
+
+        final String parentJson = extractParentFragment(type, jsonResource);
+        final String expectedParentJson = "{\"domain\":\"B\",\"scientific_name\":\"Shewanella\","
+                + "\"assembly_ref\":\"1/2/1\"}";
+        assertEquals(parentJson, expectedParentJson);
+    }
+
+    @Test
```
arfathpasha: Minor - I think the convention is to have the @Test and other annotations right above the function signature and not above the javadoc comment.

build.xml
MrCreosote: I've completely redone the test target locally.

JamesJeffryes: Seeing as there are only two test files, I'll go for the latter option

resources/typemappings/Genome.yaml
```diff
@@ -0,0 +1,27 @@
+storage-type: WS
+types:
+    KBaseGenomes.Genome:
+        default:
+           -
+               type: Genome
```
MrCreosote: Hmm, if most new genomes are going to be new style, wouldn't it be better to have the default be v2?

resources/types/FBAModel.yaml
```diff
@@ -0,0 +1,46 @@
...
+    transform: lookup.key.scientific_name
+    full-text: true
+    key-name: scientific_name
+    ui-name: Genome Name
+  - source-key: genome_guid
+    transform: lookup.key.taxonomy
+    full-text: true
+    key-name: taxonomy
+    ui-name: Genome Taxonomy
+  - source-key: genome_guid
```
MrCreosote: Is the genome id mutable?

test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -0,0 +1,518 @@
+package kbasesearchengine.test.parse;
```
MrCreosote: There's something weird here - ObjectParserTest already exists in the repo.

MrCreosote: I see the problem - it's in master, not develop.

resources/typemappings/Genome.yaml
JamesJeffryes: With the current style, wouldn't I have to list every revision from 1-12 to point them to the old style? This is much more concise and any new revisions to genome would have to be reflected here anyway.

MrCreosote:
Merge conflicts now that develop is up to date

resources/typemappings/Genome.yaml
MrCreosote: yes, but if 1-12 are listed that means if you make more changes to the genome you don't need to update the mappings unless the changes are backwards incompatible.

resources/types/FBAModel.yaml
JamesJeffryes: It's as mutable as any other field in an object I think

resources/typemappings/Genome.yaml
MrCreosote: Currently if you make even a trivial change to the the ws spec it'll get indexed as the v1 spec and fail indexing.

resources/types/FBAModel.yaml
MrCreosote: So it's immutable?

MrCreosote: Let me ask a different question - what is the genome id and where does it come from?

build.xml
```diff
@@ -155,7 +155,6 @@ java -cp ${jar.absolute.path}:${lib.classpath} kbasesearchengine.tools.SearchToo
         <sysproperty key="test.cfg" value="${test.cfg}"/>
         <batchtest>
           <fileset dir="${test.src}">
-            <include name="kbasesearchengine/test/parse/**Test.java"/>
```
MrCreosote: There were 2 tests covered by this line but only one was added below, so a test is missing

resources/types/FBAModel.yaml
JamesJeffryes: Genome ID is a string that generally comes from the user when the genome is grated by upload or an app

MrCreosote: Is it the same as the workspace object name?

JamesJeffryes: Yes it is

build.xml
JamesJeffryes: There are two, split by arfath's comment

MrCreosote: ObjectParserTests was already there. A test is missing: https://github.com/kbase/KBaseSearchEngine/tree/develop/test/src/kbasesearchengine/test/parse

resources/types/FBAModel.yaml
MrCreosote: Ok, this is something I've never understood. The ID is the object name, which can be changed, but is embedded in the object, which can't be changed. That seems incredibly confusing. I removed that from the index for Assembly (IIRC) due to this issue. Do you know why the design this way? It seems like it's bound to cause user confusion (and I know testers have pointed it out as a bug before).

JamesJeffryes: Yeah, I see your point. Is there anyway to index the object name from a guid?

MrCreosote: There is, but there's no way for that object name to be updated, so it's just as bad. The search records generated from the specs are assumed to be immutable.

MrCreosote: To be clear, the object name of the given object being indexed is also indexed and is updated with changes. If it's specified in a search spec that's not the case.

JamesJeffryes: Given those constraints, I think it's still better to leave the id in. Sure it might confuse some users if they rename a genome and it still pops up in the search but I suspect this is a much smaller issue than for users not to be able to find metabolic models generated from a given genome.

MrCreosote: Hmm, I predict bug reports in our future, so I'd like to get a call on this from @sychan or @scanon

build.xml
JamesJeffryes: Done

resources/types/FBAModel.yaml
JamesJeffryes: scanon [11:17 AM]
I think we should keep it for now.

Better to have the search option even if it sometimes would be out of sync than to totally drop it.

MrCreosote: Where is this message? I'm not able to find it in slack.

JamesJeffryes: It was a DM with Shane. We can just discuss at standup

MrCreosote:
Is this ready for re-review?

JamesJeffryes:
Yes

resources/types/FBAModel.yaml
MrCreosote: Confirmed with Shane

PR #149, opened 2018-02-06
Shutdown correctly + more

MrCreosote:
This PR has several fixes:

1) The worker and coordinator now shut down correctly when a FatalIndexingException is thrown. Headless and standard mode now use the same shutdown mechanism.
2) If the workspace failed to start up correctly a FatalIndexingException is thrown.
3) An attempt at a fix for the bizarre DuplicateKeyException error that occurs where the offending object is the value of the map, not the key. Changed a streaming implementation into a standard for loop to see if that fixes this issue. Don't understand the root cause so this may or may not work.
4) Makes the test suite fail if it can't find a test in the list of tests in build.xml rather than silently ignoring it.

Most changes occur in non-central parts of the code without extant test suites and so were tested manually.


MrCreosote:
Looking into why the coverage dropped by so much. I don't see why the misses can be +52 when the lines are only +14

MrCreosote:
For some reason it's counting tuple11 and tuple9 as a drop in coverage, even though this pr doesn't touch them... codecov is kind of goofy

lib/src/kbasesearchengine/main/SignalMonitor.java
```diff
@@ -0,0 +1,31 @@
...
+ * dormant until the signaling thread calls {@link #signal()}.
+ * 
+ * Only one signal can be sent - to send another signal, create a new class.
+ * 
+ * @author gaprice@lbl.gov
+ *
+ */
+public class SignalMonitor {
+    
+    final CountDownLatch latch = new CountDownLatch(1);
```
MrCreosote: Just noticed this should be private. Fixed locally, it'll be in a later PR.

lib/src/kbasesearchengine/main/SignalMonitor.java
```diff
@@ -0,0 +1,31 @@
...
+    
+    public SignalMonitor() {}
+    
+    public void signal() {
```
MrCreosote: also missing javadoc for these 2 methods, fixed locally.

test/src/kbasesearchengine/test/main/IndexerCoordinatorTest.java
```diff
@@ -872,4 +885,36 @@ public void logDelayedWithCacheExpiry() throws Exception {
...
+        final LineLogger logger = mock(LineLogger.class);
+        final ScheduledExecutorService executor = mock(ScheduledExecutorService.class);
+        final SignalMonitor sm = new SignalMonitor();
+        
+        final IndexerCoordinator coord = new IndexerCoordinator(storage, logger, sm, 3, executor,
+                Arrays.asList(1), ST, SC);
+        
+        when(storage.get(StatusEventProcessingState.PROC, 3))
+                .thenThrow(new FatalRetriableIndexingException("whee"));
+
```
arfathpasha: This test doesn't seem right. Looks like the 'when' clause should come above the coordinator in order to throw the exception during coord construction (I am assuming this is what you want). But even if you did that, coord would be null because of the exception thrown in the constructor and coord.awaitShutdown() would throw a null pointer.

MrCreosote: No, it gets thrown here, after construction: https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/main/IndexerCoordinator.java#L238

I purposely left it after the constructor so it wouldn't get thrown on construct.

MrCreosote: Actually, looking at this more closely, the when() isn't needed at all since the signal monitor is passed in and that's how the signal is triggered.

arfathpasha: Then you need coord.startIndexer() after the 'when' clause in order for control to reach loadEventsIntoQueue()?

MrCreosote: I just took out the when() clause and added the SignalMonitor fixes.

arfathpasha: Then test method name is misleading. This works well - 

        when(storage.get(StatusEventProcessingState.PROC, 3))
                .thenThrow(new FatalRetriableIndexingException("whee"));

        coord.startIndexer();
        Thread.sleep(100);

WIth powermokito's verifyNew, the call to the exception constructor can also be verified. But this would require powermokito's jar to be included I think.

MrCreosote: I updated the test name to reflect what's actually being tested. The reason I didn't do a full test like the above is that IIUC unit tests are supposed to mock their dependencies (other than value classes) rather than relying on dependency instances and their internal state. Admittedly, SignalMonitor is about as simple a dependency as you can get...

PR #152, opened 2018-02-07
SCT-498 Add tests for WorkspaceEventHandler.load()

MrCreosote:
... in preparation for implementing search tags retrieval in the handler.


MrCreosote:
Yeah, next time someone touches those, or when we have cycles, we should add tests

PR #155, opened 2018-02-08
SCT-560: Specs for new types

JamesJeffryes:
I took a look at the implementation and the way I'm specifying "genome_names" in the pangenome spec might work, but I'm not able to test it easily


resources/types/Pangenome.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - path: type
+    keyword-type: string
+    ui-name: Pangenome Type
+  - path: genome_refs/{size}
+    keyword-type: integer
+    key-name: genomes
+  - path: orthologs/{size}
+    keyword-type: integer
+    key-name: orthologs
+  # I expect this won't work but I'm including to spur discussion on what might
```
MrCreosote: So does this work?

resources/types/Tree.yaml
```diff
@@ -0,0 +1,16 @@
...
+versions:
+- indexing-rules:
+  - path: default_node_labels/{size}
+    keyword-type: integer
+    key-name: nodes
+    ui-name: Nodes
+  - path: type
+    keyword-type: string
+    ui-name: Tree Type
+  - path: default_node_labels/*/
```
MrCreosote: How many labels might there be? If there are thousands this might cause problems

resources/types/RNASeqSampleSet.yaml
```diff
@@ -0,0 +1,7 @@
+global-object-type: RNASeqSampleSet
+storage-type: WS
+storage-object-type: KBaseRNASeq.RNASeqSampleSet
+versions:
+- indexing-rules:
+  - path: sampleset_id
```
MrCreosote: Is this field the object name again?

resources/types/PangenomeOrthologFamily.yaml
```diff
@@ -0,0 +1,23 @@
...
+- inner-sub-type: OrthologFamily
+  path-to-sub-objects: /orthologs/*/
+  primary-key-path: id
+  indexing-rules:
+  - path: id
+    keyword-type: string
+    ui-name: ID
+  - path: function
+    full-text: true
+  - path: orthologs/[*]/0
```
MrCreosote: How many orthologs can we expect here?

resources/types/PangenomeOrthologFamily.yaml
```diff
@@ -0,0 +1,23 @@
...
+    transform: values
+    keyword-type: string
+    key-name: ortholog_gens
```
MrCreosote: ortholog_genes

resources/types/Pangenome.yaml
```diff
@@ -0,0 +1,21 @@
+global-object-type: Pangenome
```
MrCreosote: Which of these values are mutable immutables?

resources/types/RNASeqSampleSet.yaml
JamesJeffryes: Yes, it's what the people want...

resources/types/PangenomeOrthologFamily.yaml
JamesJeffryes: Depends on the number of genomes, generally ~10 maybe up to ~200

resources/types/Tree.yaml
MrCreosote: I thought that there might be a maximum size, but the ES docs don't mention one: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/array.html

So maybe it's ok. There will be a java limit of ~2B...

Of course there are indexing performance and search performance implications.

resources/types/Pangenome.yaml
JamesJeffryes: This does not work in part because of the "key-name" issue. I'm going to scratch it for this version of the specs

resources/types/Pangenome.yaml
JamesJeffryes: They are all immutable. Has separate "id" attribute that is not being indexed.

resources/types/PangenomeOrthologFamily.yaml
JamesJeffryes: Fixed (though this is currently ignored by parser)

resources/types/Pangenome.yaml
MrCreosote: I thought some of the fields being indexed were object names, which are mutable immutables in this context?

JamesJeffryes: No, none of the currently indexed fields are object names

MrCreosote: I'm confused then:

>  MrCreosote 7 hours ago Owner
> Is this field the object name again?
> JamesJeffryes 7 hours ago Member
> Yes, it's what the people want...

JamesJeffryes: That was referring to RNASequenceSet. Thought the question was specific to the Pangenome object where this comment is. 

The "default_node_labels" in the Tree object may refer to genome names which are mutable (though most will be reference genomes or feature names which are immutable)

MrCreosote: I was asking about all the types here. 

Anyway, is the RSS id and the node labels in the tree the only mutable immutables then?

Thinking further, I get the use case where you extract the id from a dependent object (like Model extracting from Genome) so you can find objects based on the dependent object name (even though that name may be out of date).

I *don't* understand the use case of indexing the object's own name via the spec. The name automatically gets indexed, so if the name isn't changed it buys nothing. If the name is changed, it just means the object always shows up in searches for its original name as well as its real name, which doesn't seem like desired behavior.

JamesJeffryes: You are right, the RSS id does not make sense alone. I've used the sample_set_description instead and added a few other indexed fields. This should make node_lables the only potentially mutable type

build.xml
```diff
@@ -155,7 +155,7 @@ java -cp ${jar.absolute.path}:${lib.classpath} kbasesearchengine.tools.SearchToo
         <classpath refid="test.classpath"/>
         <formatter type="plain" usefile="false" />
         <sysproperty key="test.cfg" value="${test.cfg}"/>
-        <test name="kbasesearchengine.test.authorization.AccessGroupCacheTest"/>
+        <!--<test name="kbasesearchengine.test.authorization.AccessGroupCacheTest"/>
```
MrCreosote: yeah... pretty sure we still want to run those tests.

resources/types/RNASeqSampleSet.yaml
```diff
@@ -0,0 +1,13 @@
+global-object-type: RNASeqSampleSet
+storage-type: WS
+storage-object-type: KBaseRNASeq.RNASeqSampleSet
+versions:
+- indexing-rules:
+  - path: sampleset_desc
+    full-text: true
+  - path: num_samples
+    key-type: intager
```
MrCreosote: integer

resources/types/RNASeqSampleSet.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - path: num_replicates
+    key-type: intager
```
MrCreosote: integer

MrCreosote:
This seems ok to me, but @arfathpasha wanted tests, so I'll pass the buck to him to merge when he's happy.

arfathpasha:
@JamesJeffryes the new tests in ElasticIndexingStorageTest look good to me. I'm thinking it would be good to move the ones in ObjectParserTest (that you created) into ElasticIndexingStorageTest so they are all in one place.

JamesJeffryes:
@arfathpasha There's already genome tests in ObjectParser. I'm thinking if we want consolidate the tests we would have to strip all this out. I don't necessarily think it's a bad thing to have different kinds of tests for the objects even though the ObjectParser tests don't do key-name changes.

PR #143, opened 2018-02-01
SCT-409 SCT-410 Decorate objects of search results with narrative info

ugswork:

PR #158, opened 2018-02-08
SCT-409 SCT-410 Add narrative decorations to returned search data

MrCreosote:
Original PR: https://github.com/kbase/KBaseSearchEngine/pull/143

For furture PRs:
* fetch user display name from the auth service
* caches for workspace and auth data


lib/src/us/kbase/common/service/Tuple5.java
```diff
@@ -0,0 +1,95 @@
+package us.kbase.common.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+
+public class Tuple5 <T1, T2, T3, T4, T5> {
+    private T1 e1;
```
arfathpasha: I have no way of knowing what each element of this tuple represents (E1, E2,etc.) has no semantics attached to them. Either document or preferably just make a POJO.

MrCreosote: The tuple class is generated by the kb-sdk compiler, so any docs would need to be added there.

That being said, there are no semantics - the semantics of each of the entries is defined by the user of the tuple. It's just a list of 5 things that can have different types.

arfathpasha: That answers a long standing question about what the heck these tuples are all about. I now see the semantics correspond to narrative_info in the spec. I guess in the ideal world we'd have a POJO with a one-one mapping to the types in the spec and this is a close approximation to that. Would be nice if kb-sdk could at least auto-document from the spec struct and documentation. I wonder if this is being considered.

MrCreosote: Note the Tuple5 could be used in multiple contexts with different semantics, so can't have exact documentation. There obviously could and should be generic documentation, but unless you, Uma, or Tian are considering it, it's not being considered at the moment :P

The tuple documentation from the spec is built into the classes that contain the tuples, but it's an unreadable mess, as you've seen.

PR #172, opened 2018-02-12
Add more diagnostics

MrCreosote:
Add a size method that doesn't depend on the memoization of the size

Add a few more printouts


MrCreosote:
@kkellerlbl @arfathpasha @sychan

PR #174, opened 2018-02-14
SCT-411 Add version

MrCreosote:


MrCreosote:
Not worth the trouble for something so trivial IMO. I would like to just delete kbase.yml but then kb-sdk test would stop working. One of the things on my very long todo list is to add some documentation delineating the files that are around just to support kb-sdk test as well as how to test without it, how to build the script, that sort of thing.

PR #175, opened 2018-02-15
[SCT-586]Add type

tiramisu24:
add object type to return

### DIFF
--- a/lib/src/kbasesearchengine/main/SearchMethods.java
+++ b/lib/src/kbasesearchengine/main/SearchMethods.java
@@ -13,6 +13,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import kbasesearchengine.AccessFilter;
 import kbasesearchengine.GetObjectsInput;
@@ -36,6 +37,7 @@ import kbasesearchengine.search.IndexingStorage;
 import kbasesearchengine.search.MatchFilter.Builder;
 import kbasesearchengine.system.IndexingRules;
 import kbasesearchengine.system.ObjectTypeParsingRules;
+import kbasesearchengine.system.SearchObjectType;
 import kbasesearchengine.system.TypeStorage;
 import us.kbase.common.service.UObject;
 
@@ -212,10 +214,15 @@ public class SearchMethods implements SearchInterface {
         if (od.getParentData().isPresent()) {
             ret.withParentData(new UObject(od.getParentData().get()));
         }
+
         ret.withObjectName(od.getObjectName().orNull());
         ret.withKeyProps(od.getKeyProperties());
         ret.withHighlight(od.getHighlight());
 
+        SearchObjectType type = od.getType().orNull();
+        if(type != null){
+            addObjectProp(ret, type.getType(), "type");
+        }
         addObjectProp(ret, od.getCreator().orNull(), "creator");
         addObjectProp(ret, od.getCopier().orNull(), "copied");
         addObjectProp(ret, od.getModule().orNull(), "module");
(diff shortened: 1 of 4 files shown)


test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -420,9 +422,113 @@ private SearchObjectsOutput searchObjects(
...
+                .put("type", "Blah")
+                .put("creator", "user")
+                .put("copied", "user2")
+                .put("module", "module")
+                .put("method", "method")
+                .put("module_ver", "2")
+                .put("commit", "commitHash")
+                .build();
+
+        Map<String, String > additonalProperties = actual.getObjectProps();
```
MrCreosote: just compare the expected map to the additionalProperties map directly

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -420,9 +422,113 @@ private SearchObjectsOutput searchObjects(
...
                 .withPostProcessing(pp)
-                .withMatchFilter(new MatchFilter().withFullTextInAll("test"))
+                .withMatchFilter(new MatchFilter().withFullTextInAll(query))
                 .withAccessFilter(new AccessFilter());
 
         return  sm.searchObjects(input, "auser");
     }
-}
+
+    @Test
```
MrCreosote: Doesn't there need to be a case where the type is null?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -420,9 +422,113 @@ private SearchObjectsOutput searchObjects(
...
+
+        assertThat("did not find objects", res1.getObjects().size() == 1, is(true));
+        kbasesearchengine.ObjectData actual = res1.getObjects().get(0);
+
+        assertThat("incorrect data", actual.getData().toString(),
+                is(new UObject("obj").toString()));
+        assertThat("incorrect guid", actual.getGuid(), is(guid.toString()));
+        Map <String, String> expectedKeyProps = new HashMap<>();
+
+        expectedKeyProps.put("key", "prop");
```
MrCreosote: Skip keys was true, so shouldn't this be empty?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: I thought this was more informative this be more informative?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: ops intent was to turn on everything

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
MrCreosote: OK. I usually go for simpler code, but I can live with a test per key.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: yes damn

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -212,16 +214,21 @@ private Pagination fromSearch(kbasesearchengine.search.Pagination pg) {
         if (od.getParentData().isPresent()) {
             ret.withParentData(new UObject(od.getParentData().get()));
         }
+
         ret.withObjectName(od.getObjectName().orNull());
         ret.withKeyProps(od.getKeyProperties());
         ret.withHighlight(od.getHighlight());
 
+        SearchObjectType type = od.getType().orNull();
+        if(type != null){
```
MrCreosote: Better practice is 
```
if (od.getType().isPresent()) {
    // do stuff
}

PR #177, opened 2018-02-15
[SCT-593] NullPointerException running search_objects query

tiramisu24:
- account for elasticsearch returning null for highlight instead of empty map
- appears to occur somewhat consistently for GUIDs and sporadically for other search queries.

### DIFF
--- a/test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
+++ b/test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
@@ -446,6 +446,20 @@ public class SearchAPIIntegrationTest {
         //highlight in objects
         assertThat("incorrect highlight", actual.getHighlight(), is(expected.getHighlight()));
 
+        //test b/c highlight is unable to find number/dates and may return null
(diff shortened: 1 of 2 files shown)


tiramisu24:
Whats the best way to test this? Can I just write a test with one of the queries that highlight happens to return null?

MrCreosote:
That's how I'd do it.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -445,6 +445,21 @@ public void highlightTest () throws Exception{
 
         //highlight in objects
         assertThat("incorrect highlight", actual.getHighlight(), is(expected.getHighlight()));
+
```
MrCreosote: This should be a unit test, not an integration test. In the interests of getting this out, do you solemnly swear to switch it to an unit test when the API changes are done?

tiramisu24: yes :)

PR #178, opened 2018-02-16
Addition - Narrative info cache and users' display names

ugswork:


MrCreosote:
I'm going to close this PR as the cache stuff should go into develop. I just wound up adapting stuff from here rather than using the whole thing.

PR #182, opened 2018-02-17
SCT-409 SCT-410 Add display names to the narrative decoration

MrCreosote:
Adapted from https://github.com/kbase/KBaseSearchEngine/pull/178

### DIFF
--- a/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
+++ b/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
@@ -15,6 +15,8 @@ import kbasesearchengine.SearchObjectsOutput;
 import kbasesearchengine.SearchTypesInput;
 import kbasesearchengine.SearchTypesOutput;
 import kbasesearchengine.TypeDescriptor;
+import kbasesearchengine.authorization.TemporaryAuth2Client;
+import kbasesearchengine.authorization.TemporaryAuth2Client.Auth2Exception;
 import kbasesearchengine.common.GUID;
 import kbasesearchengine.events.handler.WorkspaceEventHandler;
 import kbasesearchengine.tools.Utils;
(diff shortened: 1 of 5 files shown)


MrCreosote:
Merging per @sychan 

@ugswork can you review when you have time?

ugswork:
Yes, I am reviewing it.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -105,35 +115,49 @@ public GetObjectsOutput getObjects(final GetObjectsInput params, final String us
...
+            throw new JsonClientException("Failed retrieving workspace info: "
+                    + e.getMessage(), e);
+        }
+        
+        final long timeMilli = WorkspaceEventHandler.parseDateToEpochMillis(wsInfo.getE4());
+        final Tuple5<String, Long, Long, String, String> tempNarrInfo =
+                    new Tuple5<String, Long, Long, String, String>()
+                .withE3(timeMilli)      // modification time
+                .withE4(wsInfo.getE3()); // workspace user name
+                // e5 gets filled in later
```
ugswork: The real user name is not filled here, in order to combine all the userIds and make one single request to the authClient to get the display names. But since we want to cache the narrativeInfo (wsId, Tuple5<String, Long, Long, String, String>), the display name should also be updated in this method in the returned Tuple5,  so that it can be cached.

MrCreosote: > But since we want to cache the narrativeInfo (wsId, Tuple5<String, Long, Long, String, String>)

I don't think there's any requirement that the cache needs to have any particular data structure.

ugswork: There isn't any such requirement. But i felt if the narrative info for each wsId needs to be cached, then it would be cleaner, if this method is made to return all the values for the given wsID.

MrCreosote: We probably don't want to make up to 10000 network calls when 1 would do in the name of cleanliness though. Premature optimization and all that but network calls are one case where it's pretty clear cut that if they can be minimized with little effort (as is the case here) that's what should be done.

PR #188, opened 2018-02-20
Change media compound subobject key

JamesJeffryes:
This addresses the log issues like this one:
1518923196918 2018-02-18T03:06:36.918Z Error processing event for event NEW_VERSION with parent ID 5a875ac0e4b00d0e6e4d29d3: kbasesearchengine.events.exceptions.UnprocessableEventIndexingException: Could not find the subobject id for one or more of the subobjects for object WS:28269/16/8 when applying search specification MediaCompound_1

### DIFF
--- a/test/src/kbasesearchengine/test/parse/ObjectParserTest.java
+++ b/test/src/kbasesearchengine/test/parse/ObjectParserTest.java
@@ -287,7 +287,9 @@ public class ObjectParserTest {
         final List<String> expectedKeys = Arrays.asList("inchikey", "concentration",
                 "minFlux", "maxFlux", "id", "name", "compound_ref");
         final List<Double> expectedFlux = Arrays.asList( 100.0, 10.0, 50.0);
+        final List<String> names = Arrays.asList( "cpd00027", "cpd00013", "cpd00009");
         final Double expectedConcentration = 0.001;
+        final String inchikey = "WQZGKKKJIJFFOK-VFUOTHLCSA-N";
         int value_index = 0;
         for (String jsonString : guidToJson.values()) {
             @SuppressWarnings("unchecked")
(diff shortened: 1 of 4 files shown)


test/src/kbasesearchengine/test/parse/ObjectParserTest.java
```diff
@@ -283,7 +285,7 @@ public void parseSubObjectsMediaCompoundTest() throws Exception {
         }
 
         final List<String> expectedKeys = Arrays.asList("inchikey", "concentration",
-                "minFlux", "maxFlux", "id", "name");
```
arfathpasha: Not a big deal, but for an exact match, expectedKeys = Arrays.asList("maxFlux", "minFlux", "name", "concentration", "id", "compound_ref")

JamesJeffryes: Is just checking the key like this ok? 
assertThat(expectedKeys.containsAll(result.keySet()), is(true));

arfathpasha: Sorry @JamesJeffryes I missed your question in a sea of pr emails. I think what you have is OK to get stuff out now, but if you are asking if the check is sufficient, I think it can be made more complete. To do this, your test media01.json should include all fields (required and optional), then you can check to ensure result.keySet().size() is the right size and check if the values of each of the keys is what you expect.

JamesJeffryes:
@MrCreosote I went ahead and made the change he requested to my test. Should be good to merge

MrCreosote:
@JamesJeffryes 

> and check if the values of each of the keys is what you expect.

It looks like the tests don't do this unless I'm misunderstanding them - it looks like only a subset of the keys are checked.

MrCreosote:
@arfathpasha as far as I can tell this now meets what you asked for, but if not feel free to hound James into madness next week to get it done.

PR #189, opened 2018-02-20
Fix coordinator load bug

MrCreosote:
The coordinator currently has no way to know which events in the DB are in memory without passing all the event IDs into the DB query, and so continually loads the same UNPROC events over and over, duplicating queued events and eventually filling the queue. 

This change makes the queue ignore events that already exist in the queue based on the event id. The same UNPROC events will get pulled from the DB over and over, but will not be added to the queue. As before, events that have been set to READY will no longer get pulled.

This does mean that the effective load size will be less than requested, since some of the events will be ignored. Since the oldest events will still get loaded before newer events, things should work ok.

A pathological example might be if 5000 events are in the queue, and therefore the coordinator requests a load of 5000 events. The same 5000 events may be loaded and all ignored if no changes have occurred in the DB (e.g. the creation of events with older timestamps than the 5000). Eventually the coordinator will work through the 5000 events, setting them to READY, and the load will no longer include those events. If a new event is added to the DB with an older time stamp, it will take the place of one of the 5000 events in the next load.

Longer term, and if performance becomes an issue (which seems unlikely), the coordinator could mark events as loaded in the db and not include them in the db query results. Those marks would need to be cleared on coordinator startup.

### DIFF
--- a/lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
+++ b/lib/src/kbasesearchengine/events/AccessGroupEventQueue.java
@@ -56,6 +56,9 @@ public class AccessGroupEventQueue {
     private StoredStatusEvent ready = null; // AG event ready for processing
     private StoredStatusEvent processing = null; // AG event processing
     private int size = 0; // keep a record of size rather than running through sub queues
+    // this contains only access group level events. Object level events are handled by their
+    // respective queues.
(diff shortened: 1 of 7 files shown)


MrCreosote:
I wonder if the UNPROC events really need to be kept in memory - maybe the UNPROC events can just be pulled each time and the READY and PROC events kept in memory. Longer term change for now.

PR #190, opened 2018-02-20
speculative changes to support filtering

eapearson:
- currently the  only test I've been doing, IndexerWorderIntegrationTest

- the last two narrative search tests fail.  
checkSearch(1, ImmutableList.of("Narrative"), "markdown", wsid, true);
checkSearch(1, ImmutableList.of("Narrative"), "code", wsid, true);

They are testing searching against the cell_type, and poking around it looks like the values are maybe not being collected (although the other search tests don't fail, but don't know if they are testing the values transform.)

maybe I broke something in the array handling, will look at when have a chance today.


MrCreosote:
per @scanon, we're going to do something simpler for now and work on a way to run custom code for indexing stuff like this.

PR #200, opened 2018-02-23
Revert stream to regular map construction

MrCreosote:
Getting weird errors:

java.lang.IllegalStateException: Duplicate key kbasesearchengine.search.ObjectData@2ab1c576

Note that ObjectData is the *value*


MrCreosote:
Urrg this went to master instead of develop

PR #209, opened 2018-02-26
SCT-626 Refactor elastic client usage

MrCreosote:
All client usage now occurs in one method.

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -367,7 +367,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
                 pw.println(UObject.transformObjectToString(doc));
             }
             pw.close();
-            makeBulkRequest("POST", indexName, tempFile);
+            makeRequestBulk("POST", indexName, tempFile);
             updateLastVersionsInData(indexName, pguid, lastVersion);
         } finally {
             tempFile.delete();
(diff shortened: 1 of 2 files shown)


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1691,25 +1686,44 @@ public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder hacb) {
...
     
-    public Response makeRequest(String reqType, String urlPath, Map<String, ?> doc, 
-            Map<String, String> attributes) throws IOException {
+    private Response makeRequest(
+            final String reqType,
+            final String urlPath,
+            final Map<String, ?> doc, 
+            final Map<String, String> attributes)
+            throws IOException {
+        return makeRequest(reqType, urlPath, attributes, doc == null ? null : stringEntity(
```
arfathpasha: (HttpEntity)stringEntity(
                UObject.transformObjectToString(doc))

MrCreosote: Why is the cast necessary?

MrCreosote:
I'm going to merge since I'm ready for the next PR.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Helps the reader match the arguments to the right function signature.

MrCreosote: So you're advocating for including a cast every time a polymorphic method is called and the type is unclear?

arfathpasha: My mistake. My IDE was not making the connection to the function signature and it was an error with the way I manually included the jars. Please ignore.

PR #210, opened 2018-02-27
SCT-626 Treat elastic conflict exceptions as non-fatal

MrCreosote:
They're not great in that they mean the code isn't handling indexing
correctly in some cases, but shutting the worker down and leaving events
in a stuck state is worse.

This doesn't fix the root cause of the problem but means there's no
known errors that cause a worker shutdown where the worker continuing
does no further harm.

Unfortunately reliably testing elastic conflicts is going to be very difficult (at least I can't think of a simple way to do it) so we'll have to test in CI.

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -1700,7 +1700,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
         try {
             return makeRequest(reqType, urlPath, doc);
         } catch (IndexingConflictException e) {
-            // this is impossible to test, and so is not tested
+            // this is very difficult to test, and so is not tested
             throw new IOException(
                     "This operation is not expected to result in a conflict, yet it occurred: " +
                     e.getMessage(), e);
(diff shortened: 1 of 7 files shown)


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1721,11 +1741,15 @@ private Response makeRequest(
...
             final HttpEntity body)
-            throws IOException {
+            throws IOException, IndexingConflictException {
         try {
             return getRestClient().performRequest(reqType, urlPath, attributes, body);
-        } catch (ResponseException ex) {
-            throw new IOException(ex.getMessage(), ex);
+        } catch (ResponseException re) {
+            if (re.getResponse().getStatusLine().getStatusCode() == 409) {
+                // this is really difficult to test, and so is not
```
arfathpasha: comment looks incomplete? and so is not...

MrCreosote: fixed

MrCreosote:
Confirmed working in CI

PR #212, opened 2018-02-28
SCT-642 Add error type to errors

MrCreosote:
Adds an error type to errors that reach the portion of the worker code
where events get updated in the DB. Currently does nothing, but future
PRs will add the error type to the DB on a failed event. This will allow
for filtering on the error type when doing error analysis, rather than
having to filter on arbitrary error strings.

### DIFF
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -43,6 +43,7 @@ import kbasesearchengine.events.handler.EventHandler;
 import kbasesearchengine.events.handler.ResolvedReference;
 import kbasesearchengine.events.handler.SourceData;
 import kbasesearchengine.events.storage.StatusEventStorage;
+import kbasesearchengine.parse.ContigLocationException;
 import kbasesearchengine.parse.KeywordParser;
 import kbasesearchengine.parse.ObjectParseException;
 import kbasesearchengine.parse.ObjectParser;
@@ -631,8 +632,10 @@ public class IndexerWorker implements Stoppable {
              * File IO problems are generally going to mean something is very wrong
              * (like bad disk), since the file should already exist at this point.
              */
+        } catch (ContigLocationException e) {
+            throw new UnprocessableEventIndexingException(
+                    ErrorType.LOCATION_ERROR, e.getMessage(), e);
         } catch (ObjectParseException e) {
-            //TODO NNOW location exception
             throw new UnprocessableEventIndexingException(ErrorType.OTHER, e.getMessage(), e);
         } catch (IOException e) {
             throw new FatalRetriableIndexingException(ErrorType.OTHER, e.getMessage(), e);
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -143,11 +143,11 @@ public class IndexerWorker implements Stoppable {
                     // keep processing events until there are none left
                     processedEvent = performOneTick();
                 } catch (InterruptedException | FatalIndexingException e) {
-                    logError(ErrorDisp.FATAL, e);
+                    logError(LogPrefix.FATAL, e);
                     executor.shutdown();
                     signalMonitor.signal();
                 } catch (Throwable e) {
-                    logError(ErrorDisp.UNEXPECTED, e);
+                    logError(LogPrefix.UNEXPECTED, e);
                 }
             }
         }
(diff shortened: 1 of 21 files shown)


lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -169,18 +171,19 @@ public void stop(long millisToWait) throws InterruptedException {
...
         Utils.nonNull(errtype, "errtype");
         final String msg;
-        if (ErrorType.FATAL.equals(errtype)) {
+        if (ErrorDisp.FATAL.equals(errtype)) {
             msg = "Fatal error in indexer, shutting down";
-        } else if (ErrorType.STD.equals(errtype)) {
+        } else if (ErrorDisp.STD.equals(errtype)) {
             msg = "Error in indexer";
-        } else if (ErrorType.UNEXPECTED.equals(errtype)) {
+        } else if (ErrorDisp.UNEXPECTED.equals(errtype)) {
```
arfathpasha: Why not capture error code and message in the enum itself rather than having the two in separate places. 

```
public enum Error {
  FATAL(0, "Fatal error in indexer."),
  STD(1, "Standard error in indexer.");

  private final int code;
  private final String description;

  private Error(int code, String description) {
    this.code = code;
    this.description = description;
  }

  public String getDescription() {
     return description;
  }

  public int getCode() {
     return code;
  }

  @Override
  public String toString() {
    return code + ": " + description;
  }
}
```

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -169,18 +171,19 @@ public void stop(long millisToWait) throws InterruptedException {
         }
     }
     
-    private enum ErrorType {
+    // error disposition
+    private enum ErrorDisp {
```
arfathpasha: How about consolidating all errors into one big enum,  or even better, externalize the errors in a properties file and read them into the enum or class. This will make it easier to add to documentation, and the ops folks can access easily.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: That's probably better, but this is out of scope for this PR. This enum is just a ternary selector for a general log error prefix string - if there were only two possibilities I would've just used a boolean. The only reason it shows up in this PR is because I renamed the enum to prevent a name conflict with the error type enum.

I'm not sure what the point of the code is - these aren't error types per se, they're just a ternary.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: This enum and the ErrorType enum are entirely orthogonal - as I noted above, one is just a ternary selector for a log prefix. The new ErrorType enum actually contains error types.

For the ErrorType enum, it's not clear to me how loading the contents from a file would work, since the code selects specific known error codes based on various scenarios. How is would pick an error code from an arbitrary list of codes isn't clear.

In any case, IMO there are much larger priorities and we should wait for a request from ops before we implement something they may or may not want.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I guess it's OK. But there is another private enum ErrorType in IndexerCoordinator that you may want to rename as ErrorDisp for consistency. To me they feel like error types, have the same values, and since they are being used in more than one place, I thought from a code quality perspective they could use some consolidation into an n-ary set of types with code and message as you have in ErrorType.java (only the message is in the form of a comment there).

MrCreosote:
Need to watch out for PRs going against master instead of develop

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Something like,
```
LOCATION_ERROR=Lack of location data for a location transform.
SUBOBJECT_COUNT=An object has too many subobjects to index. 
...
```
in a properties file that can be referenced from an ops manual in their entirety. 
Then, 
```
Enum ErrorTypes {
   properties = new Properties();

...
   try {
           properties.load(Constants.class.getResourceAsStream(PATH));
   }
...

public void getMessage(String type) {
    properties.get(type);  // returns message or throws error if type does not exist.
}
}
```

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: > But there is another private enum ErrorType in IndexerCoordinator that you may want to rename as ErrorDisp for consistency.

I forgot about that one. That one only had 2 states so I just got rid of it and used a boolean.

When I was writing the code, the enums were a) internal only and b) so trivially simple that I didn't worry about DRYing them out. 

I've refactored the old ErrorType class to better reflect its functionality.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: But what is the point of that string? It wouldn't be used anywhere.

arfathpasha: Serves as a key/code/reference. Makes it easy when talking about the error.

MrCreosote: But if it's never used in the code, it won't serve that function. I'm only planning to stick the enum values into the DB, so any strings associated with those values will never see the light of day.

arfathpasha: OK. I was thinking along the lines of having error codes that could be used in log messages and exceptions with which logs can be easily searched etc. since you had once mentioned that you wanted that feature. But this is fine.

MrCreosote: No, the point of all this work is to *avoid* having to go to the logs. Hopefully once it's done the logs will only be needed for getting the entire stack trace for truncated traces (which will be easy to find by grepping for event ID) and for looking at retry rates, since the retries obviously won't be going in the DB. I don't see the latter happening very often.

PR #214, opened 2018-03-02
SCT-624 Simplify object type supported check

MrCreosote:
listObjectTypeParsingRules no longer throws any checked exceptions so
wrapping it in a retrier is useless. Also relocate the method near to
the one place it's used.

Trying to simplify the exception handling as much as possible before adding the exceptions to the DB.


MrCreosote:
I guess we don't have any tests for type skipping. I'll make some changes to make this testable and add some tests in my next PR.

PR #216, opened 2018-03-03
Add test for event skipping

MrCreosote:


MrCreosote:
Well, reran the build and codecov is still not picking up the report, not sure what else to do here. Should be 100% coverage though.

MrCreosote:
That's right, I accidentally pushed some commits up after you approved the request and force pushed to reset the branch. I guess that confuses codecov.

PR #217, opened 2018-03-08
SCT-649: Index taxons

JamesJeffryes:
**This PR builds on SCT-560. That PR should be merged first in order to simplify review of these changes.**

This adds an index rule for the Taxon type to support use of the Search code for assigning Taxonomy to genomes per Shane's request. One thing to note is that it's probably undesirable that that taxons show up in general user search. We should look into what would be necessary to blacklist the type or workspace.


JamesJeffryes:
Added the requested tests

PR #221, opened 2018-03-14
SCT-624 Store indexing exceptions in DB

MrCreosote:
Sets up the glue code that connects all the changes in previous PRs (e.g. new exception types & error codes, error storage with events) such that indexing errors are now stored in the database along with their associated events.

### DIFF
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -142,7 +142,7 @@ public class IndexerWorker implements Stoppable {
                 processedEvent = false;
                 try {
                     // keep processing events until there are none left
-                    processedEvent = performOneTick();
+                    processedEvent = runCycle();
                 } catch (InterruptedException | FatalIndexingException e) {
                     logError(LogPrefix.FATAL, e);
                     executor.shutdown();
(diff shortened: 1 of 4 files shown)


lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -213,25 +213,41 @@ private void logError(
...
         boolean processedEvent = false;
         if (optEvent.isPresent()) {
             final StoredStatusEvent parentEvent = optEvent.get();
             final EventHandler handler;
             try {
                 handler = getEventHandler(parentEvent);
             } catch (UnprocessableEventIndexingException e) {
-                logError(LogPrefix.STD, e);
-                markEventProcessed(parentEvent, StatusEventProcessingState.FAIL);
+                handleException("Error getting event handler", parentEvent, e);
```
arfathpasha: looks like handleException sets the event state as PROC. Shouldn't it be FAIL in this case?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -213,25 +213,41 @@ private void logError(
...
                 return true;
```
arfathpasha: Shouldn't this line return false?

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: No, the PROC tells the method which event state is allowed when looking up the event. It's probably not really needed but doesn't hurt.

MrCreosote: The method that takes an exception as an argument when setting the state always sets the state to FAIL.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: No, the boolean tells the caller whether an event was processed or not, regardless of the final state. The purpose of the boolean is so the calling loop can continue to pull events until there are none left with no delay between events. When this function returns false, the loop will wait.

PR #222, opened 2018-03-14
SCT-667: Updates to Genome indexing

JamesJeffryes:
Also SCT-668

This PR addresses Eric's index update requests:
https://docs.google.com/document/d/1t53t8P0YtJzJXy8sPJ8IUxbpoiEZcXRH6swAS-q9Tjg/edit#


arfathpasha:
Looks like you missed one keyword: string?

https://github.com/kbase/KBaseSearchEngine/blob/master/resources/types/AssemblyContig.yaml#L11

Also, we need genome's scientific name to be indexed as both full-text and keyword, so I suggest adding a new field as follows - 
```
                # retain existing field
                path: scientific_name
                full-text: true
                
                # add this new field
                path: scientific_name_keyword
                keyword-type: keyword
```

PR #231, opened 2018-03-20
SCT-688 Refactor API object properties

MrCreosote:
1) Object properties are always returned in practice, so remove
object_info boolean.

2) Move the object properties to the root of the data structure for ease
of use of the API.

KBase SDK version 1.0.18 (commit
e29f39ed467111aa60bdca74b3d45439d76e71c2)


MrCreosote:
@eapearson

PR #238, opened 2018-04-25
SCT-753 Remove recursion from indexing

arfathpasha:
- Modularized new code into separate function calls
- Verified in IndexerWorkerIntegration test that assembly and assembly contigs are not recursively indexed
- Verified using FBAModel in IndexerWorkerIntegration test that lookup transforms recursively extract data from referenced objects via workspace 
- I thought about removing the use of ObjectData between the lookup provider and keyword parser but that would only make the code more ugly.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,7 +3,8 @@ Search Service MKII release notes
 
 0.2.0
 -----
-1. Removed recursive indexing. 
+1. Removed recursive indexing to prevent race conditions during indexing (this change
+   has no impact on existing indexes).
 
 0.1.1
 -----
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,10 @@
 Search Service MKII release notes
 =================================
 
+0.2.0
+-----
+1. Removed recursive indexing. 
+
 0.1.1
 -----
 
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -820,7 +820,9 @@ public class IndexerWorker implements Stoppable {
             pp.objectData = false;
             pp.objectKeys = true;
 
-            List<ObjectData> data = new ArrayList<>();
+            final List<ObjectData> data = new ArrayList<>();
+
+            // get object from data source
             for(GUID guid: guids) {
                 File tempFile = null;
                 try {
@@ -831,75 +833,98 @@ public class IndexerWorker implements Stoppable {
                 }
 
                 try {
-                    // make a copy to avoid mutating the caller's path
-                    final LinkedList<GUID> newRefPath = new LinkedList<>();
-                    newRefPath.add(guid);
-
                     final EventHandler handler = getEventHandler(guid);
                     Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
                     while(refs.hasNext()) {
                         ResolvedReference ref = refs.next();
-                        final SourceData obj = handler.load(newRefPath, tempFile.toPath());
-                        final List<ObjectTypeParsingRules> parsingRules = new ArrayList<>(
-                                typeStorage.listObjectTypeParsingRules(ref.getType()));
 
-                        // filter by subObject type
-                        final List<ObjectTypeParsingRules> filteredParsingRules = new ArrayList<>();
-                        for (final ObjectTypeParsingRules rule : parsingRules) {
-                            String refSubObjectType = ref.getReference().getSubObjectType();
-                            String ruleSubObjectType = rule.getSubObjectType().orNull();
-                            if(refSubObjectType == null && ruleSubObjectType == null) {
-                                filteredParsingRules.add(rule);
-                            }
-                            else if(refSubObjectType != null &&
-                                    refSubObjectType.equals(ruleSubObjectType)) {
-                                filteredParsingRules.add(rule);
-                            }
-                        }
-                        Collections.sort(filteredParsingRules, new ParsingRulesSubtypeFirstComparator());
-                        //List<ObjectData> data = new ArrayList<>();
+                        // make a copy to avoid mutating the caller's path
+                        final LinkedList<GUID> newRefPath = new LinkedList<>();
+                        newRefPath.add(guid);
+                        final SourceData obj = handler.load(newRefPath, tempFile.toPath());
+
+                        List<ObjectTypeParsingRules> filteredParsingRules = getFilteredSortedParsingRules(ref);
+
                         for (final ObjectTypeParsingRules rule : filteredParsingRules) {
                             final ParseObjectsRet parsedRet = parseObjects(guid, this,
                                     newRefPath, obj, rule);
+
                             GUID parsedGUID = parsedRet.guidToObj.keySet().iterator().next();
-                            ObjectData.Builder objDataBuilder = ObjectData.getBuilder(parsedGUID,rule.getGlobalObjectType());
-                            Map<String, List<Object>> keywords = parsedRet.guidToObj.get(parsedGUID).getKeywords();
-                            Iterator<String> keys = keywords.keySet().iterator();
-                            while(keys.hasNext()) {
-                                String key = keys.next();
-                                String textValue;
-                                Object objValue = keywords.get(key);
-                                //////////////////////////////////////////////
-                                if (objValue instanceof List) {
-                                    @SuppressWarnings("unchecked")
-                                    final List<Object> objValue2 = (List<Object>) objValue;
-                                    textValue = objValue2.stream().map(Object::toString)
-                                            .collect(Collectors.joining(", "));
-                                } else {
-                                    textValue = String.valueOf(objValue);
-                                }
-                                //////////////////////////////////////////////
-                                objDataBuilder.withKeyProperty(key, textValue);
-                            }
-                            objDataBuilder.withNullableObjectName(obj.getName());
-                            objDataBuilder.withNullableCreator(obj.getCreator());
-                            objDataBuilder.withNullableMD5(obj.getCommitHash().orNull());
-                            data.add(objDataBuilder.build());
+
+                            ObjectData objData = buildObjectDataFrom(
+                                    parsedRet.guidToObj.keySet().iterator().next(),
+                                    rule.getGlobalObjectType(),
+                                    obj.getName(),
+                                    obj.getCreator(),
+                                    obj.getCommitHash(),
+                                    parsedRet.guidToObj.get(parsedGUID).getKeywords());
+
+                            data.add(objData);
                         }
                     }
                 } catch(Exception ex) {
                     ex.printStackTrace();
                 }
-
             }
             return data;
-//            try {
-//                List<ObjectData> data2 = indexingStorage.getObjectsByIds(guids, pp);
-//                return data2;
-//            } catch (IOException e) {
-//                throw new RetriableIndexingException(ErrorType.OTHER, e.getMessage(), e);
-//            }
-                //////////////////////////////////////////////////////////////////
+        }
+
+        private List<ObjectTypeParsingRules> getFilteredSortedParsingRules(ResolvedReference ref) {
+
+            final List<ObjectTypeParsingRules> parsingRules = new ArrayList<>(
+                    typeStorage.listObjectTypeParsingRules(ref.getType()));
+
+            // filter by subObject type
+            final List<ObjectTypeParsingRules> filteredParsingRules = new ArrayList<>();
+            for (final ObjectTypeParsingRules rule : parsingRules) {
+                String refSubObjectType = ref.getReference().getSubObjectType();
+                String ruleSubObjectType = rule.getSubObjectType().orNull();
+                if(refSubObjectType == null && ruleSubObjectType == null) {
+                    filteredParsingRules.add(rule);
+                }
+                else if(refSubObjectType != null &&
+                        refSubObjectType.equals(ruleSubObjectType)) {
+                    filteredParsingRules.add(rule);
+                }
+            }
+
+            // sort
+            Collections.sort(filteredParsingRules, new ParsingRulesSubtypeFirstComparator());
+
+            return filteredParsingRules;
+        }
+
+
+
+        private ObjectData buildObjectDataFrom(GUID guid,
+                                               SearchObjectType globalObjectType,
+                                               String objectName,
+                                               String objectCreator,
+                                               Optional<String> md5,
+                                               Map<String, List<Object>> keywords) {
+
+            ObjectData.Builder objDataBuilder = ObjectData.getBuilder(guid,globalObjectType);
+            Iterator<String> keys = keywords.keySet().iterator();
+            while(keys.hasNext()) {
+                String key = keys.next();
+                String textValue;
+                Object objValue = keywords.get(key);
+
+                if (objValue instanceof List) {
+                    @SuppressWarnings("unchecked")
+                    final List<Object> objValue2 = (List<Object>) objValue;
+                    textValue = objValue2.stream().map(Object::toString)
+                            .collect(Collectors.joining(", "));
+                } else {
+                    textValue = String.valueOf(objValue);
+                }
+
+                objDataBuilder.withKeyProperty(key, textValue);
+            }
+            objDataBuilder.withNullableObjectName(objectName);
+            objDataBuilder.withNullableCreator(objectCreator);
+            objDataBuilder.withNullableMD5(md5.orNull());
+            return objDataBuilder.build();
         }
         
         @Override
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -749,10 +749,6 @@ public class IndexerWorker implements Stoppable {
                     final GUID guid = rr.getResolvedReference();
                     final boolean indexed = retrier.retryFunc(
                             g -> checkParentGuidExists(g), guid, null);
-//                    if (!indexed) {
-//                        indexObjectWrapperFn(guid, rr.getType(), rr.getTimestamp(), false,
-//                                this, callerRefPath);
-//                    }
                     ret.add(guid);
                     refResolvingCache.get(storageCode)
                             .put(refToRefPath.get(rr.getReference()), guid);
@@ -789,32 +785,6 @@ public class IndexerWorker implements Stoppable {
 
             return eh.resolveReferences(callerRefPath, refsToResolve);
         }
-        
-//        private void indexObjectWrapperFn(
-//                final GUID guid,
-//                final StorageObjectType storageObjectType,
-//                final Instant timestamp,
-//                final boolean isPublic,
-//                final ObjectLookupProvider indexLookup,
-//                final List<GUID> objectRefPath)
-//                throws IndexingException, InterruptedException {
-//            final List<Object> input = Arrays.asList(guid, storageObjectType, timestamp, isPublic,
-//                    indexLookup, objectRefPath);
-//            retrier.retryCons(i -> indexObjectWrapperFn(i), input, null);
-//        }
-
-//        private void indexObjectWrapperFn(final List<Object> input)
-//                throws IndexingException, InterruptedException, RetriableIndexingException {
-//            final GUID guid = (GUID) input.get(0);
-//            final StorageObjectType storageObjectType = (StorageObjectType) input.get(1);
-//            final Instant timestamp = (Instant) input.get(2);
-//            final boolean isPublic = (boolean) input.get(3);
-//            final ObjectLookupProvider indexLookup = (ObjectLookupProvider) input.get(4);
-//            @SuppressWarnings("unchecked")
-//            final List<GUID> objectRefPath = (List<GUID>) input.get(5);
-//
-//            indexObject(guid, storageObjectType, timestamp, isPublic, indexLookup, objectRefPath);
-//        }
 
         @Override
         public Map<GUID, ObjectData> lookupObjectsByGuid(final Set<GUID> guids)
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -815,16 +815,12 @@ public class IndexerWorker implements Stoppable {
         
         private List<ObjectData> getObjectsByIds(final Set<GUID> guids)
                 throws RetriableIndexingException {
-            kbasesearchengine.search.PostProcessing pp = 
-                    new kbasesearchengine.search.PostProcessing();
-            pp.objectData = false;
-            pp.objectKeys = true;
 
             final List<ObjectData> data = new ArrayList<>();
 
             // get object from data source
             for (GUID guid: guids) {
-                File tempFile = null;
+                final File tempFile;
                 try {
                     FileUtil.getOrCreateSubDir(rootTempDir, guid.getStorageCode());
                     tempFile = File.createTempFile("ws_srv_response_", ".json");
@@ -834,24 +830,24 @@ public class IndexerWorker implements Stoppable {
 
                 try {
                     final EventHandler handler = getEventHandler(guid);
-                    Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
+                    final Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
                     while (refs.hasNext()) {
-                        ResolvedReference ref = refs.next();
+                        final ResolvedReference ref = refs.next();
 
                         // make a copy to avoid mutating the caller's path
                         final LinkedList<GUID> newRefPath = new LinkedList<>();
                         newRefPath.add(guid);
                         final SourceData obj = handler.load(newRefPath, tempFile.toPath());
 
-                        List<ObjectTypeParsingRules> filteredParsingRules = getFilteredSortedParsingRules(ref);
+                        final List<ObjectTypeParsingRules> filteredParsingRules = getFilteredSortedParsingRules(ref);
 
                         for (final ObjectTypeParsingRules rule : filteredParsingRules) {
                             final ParseObjectsRet parsedRet = parseObjects(guid, this,
                                     newRefPath, obj, rule);
 
-                            GUID parsedGUID = parsedRet.guidToObj.keySet().iterator().next();
+                            final GUID parsedGUID = parsedRet.guidToObj.keySet().iterator().next();
 
-                            ObjectData objData = buildObjectDataFrom(
+                            final ObjectData objData = buildObjectDataFrom(
                                     parsedRet.guidToObj.keySet().iterator().next(),
                                     rule.getGlobalObjectType(),
                                     obj.getName(),
@@ -877,8 +873,8 @@ public class IndexerWorker implements Stoppable {
             // filter by subObject type
             final List<ObjectTypeParsingRules> filteredParsingRules = new ArrayList<>();
             for (final ObjectTypeParsingRules rule : parsingRules) {
-                String refSubObjectType = ref.getReference().getSubObjectType();
-                String ruleSubObjectType = rule.getSubObjectType().orNull();
+                final String refSubObjectType = ref.getReference().getSubObjectType();
+                final String ruleSubObjectType = rule.getSubObjectType().orNull();
                 if (refSubObjectType == null && ruleSubObjectType == null) {
                     filteredParsingRules.add(rule);
                 }
@@ -906,9 +902,9 @@ public class IndexerWorker implements Stoppable {
             ObjectData.Builder objDataBuilder = ObjectData.getBuilder(guid,globalObjectType);
             Iterator<String> keys = keywords.keySet().iterator();
             while (keys.hasNext()) {
-                String key = keys.next();
-                String textValue;
-                Object objValue = keywords.get(key);
+                final String key = keys.next();
+                final String textValue;
+                final Object objValue = keywords.get(key);
 
                 if (objValue instanceof List) {
                     @SuppressWarnings("unchecked")
--- a/lib/src/kbasesearchengine/main/IndexerWorker.java
+++ b/lib/src/kbasesearchengine/main/IndexerWorker.java
@@ -19,6 +19,7 @@ import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.google.common.base.Optional;
@@ -748,10 +749,10 @@ public class IndexerWorker implements Stoppable {
                     final GUID guid = rr.getResolvedReference();
                     final boolean indexed = retrier.retryFunc(
                             g -> checkParentGuidExists(g), guid, null);
-                    if (!indexed) {
-                        indexObjectWrapperFn(guid, rr.getType(), rr.getTimestamp(), false,
-                                this, callerRefPath);
-                    }
+//                    if (!indexed) {
+//                        indexObjectWrapperFn(guid, rr.getType(), rr.getTimestamp(), false,
+//                                this, callerRefPath);
+//                    }
                     ret.add(guid);
                     refResolvingCache.get(storageCode)
                             .put(refToRefPath.get(rr.getReference()), guid);
@@ -789,31 +790,31 @@ public class IndexerWorker implements Stoppable {
             return eh.resolveReferences(callerRefPath, refsToResolve);
         }
         
-        private void indexObjectWrapperFn(
-                final GUID guid,
-                final StorageObjectType storageObjectType,
-                final Instant timestamp,
-                final boolean isPublic,
-                final ObjectLookupProvider indexLookup,
-                final List<GUID> objectRefPath) 
-                throws IndexingException, InterruptedException {
-            final List<Object> input = Arrays.asList(guid, storageObjectType, timestamp, isPublic,
-                    indexLookup, objectRefPath);
-            retrier.retryCons(i -> indexObjectWrapperFn(i), input, null);
-        }
+//        private void indexObjectWrapperFn(
+//                final GUID guid,
+//                final StorageObjectType storageObjectType,
+//                final Instant timestamp,
+//                final boolean isPublic,
+//                final ObjectLookupProvider indexLookup,
+//                final List<GUID> objectRefPath)
+//                throws IndexingException, InterruptedException {
+//            final List<Object> input = Arrays.asList(guid, storageObjectType, timestamp, isPublic,
+//                    indexLookup, objectRefPath);
+//            retrier.retryCons(i -> indexObjectWrapperFn(i), input, null);
+//        }
 
-        private void indexObjectWrapperFn(final List<Object> input)
-                throws IndexingException, InterruptedException, RetriableIndexingException {
-            final GUID guid = (GUID) input.get(0);
-            final StorageObjectType storageObjectType = (StorageObjectType) input.get(1);
-            final Instant timestamp = (Instant) input.get(2);
-            final boolean isPublic = (boolean) input.get(3);
-            final ObjectLookupProvider indexLookup = (ObjectLookupProvider) input.get(4);
-            @SuppressWarnings("unchecked")
-            final List<GUID> objectRefPath = (List<GUID>) input.get(5);
-            
-            indexObject(guid, storageObjectType, timestamp, isPublic, indexLookup, objectRefPath);
-        }
+//        private void indexObjectWrapperFn(final List<Object> input)
+//                throws IndexingException, InterruptedException, RetriableIndexingException {
+//            final GUID guid = (GUID) input.get(0);
+//            final StorageObjectType storageObjectType = (StorageObjectType) input.get(1);
+//            final Instant timestamp = (Instant) input.get(2);
+//            final boolean isPublic = (boolean) input.get(3);
+//            final ObjectLookupProvider indexLookup = (ObjectLookupProvider) input.get(4);
+//            @SuppressWarnings("unchecked")
+//            final List<GUID> objectRefPath = (List<GUID>) input.get(5);
+//
+//            indexObject(guid, storageObjectType, timestamp, isPublic, indexLookup, objectRefPath);
+//        }
 
         @Override
         public Map<GUID, ObjectData> lookupObjectsByGuid(final Set<GUID> guids)
@@ -848,11 +849,87 @@ public class IndexerWorker implements Stoppable {
                     new kbasesearchengine.search.PostProcessing();
             pp.objectData = false;
             pp.objectKeys = true;
-            try {
-                return indexingStorage.getObjectsByIds(guids, pp);
-            } catch (IOException e) {
-                throw new RetriableIndexingException(ErrorType.OTHER, e.getMessage(), e);
+
+            List<ObjectData> data = new ArrayList<>();
+            for(GUID guid: guids) {
+                File tempFile = null;
+                try {
+                    FileUtil.getOrCreateSubDir(rootTempDir, guid.getStorageCode());
+                    tempFile = File.createTempFile("ws_srv_response_", ".json");
+                } catch (IOException ex) {
+                    throw new RetriableIndexingException(ErrorType.OTHER, ex.getMessage(), ex);
+                }
+
+                try {
+                    // make a copy to avoid mutating the caller's path
+                    final LinkedList<GUID> newRefPath = new LinkedList<>();
+                    newRefPath.add(guid);
+
+                    final EventHandler handler = getEventHandler(guid);
+                    Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
+                    while(refs.hasNext()) {
+                        ResolvedReference ref = refs.next();
+                        final SourceData obj = handler.load(newRefPath, tempFile.toPath());
+                        final List<ObjectTypeParsingRules> parsingRules = new ArrayList<>(
+                                typeStorage.listObjectTypeParsingRules(ref.getType()));
+
+                        // filter by subObject type
+                        final List<ObjectTypeParsingRules> filteredParsingRules = new ArrayList<>();
+                        for (final ObjectTypeParsingRules rule : parsingRules) {
+                            String refSubObjectType = ref.getReference().getSubObjectType();
+                            String ruleSubObjectType = rule.getSubObjectType().orNull();
+                            if(refSubObjectType == null && ruleSubObjectType == null) {
+                                filteredParsingRules.add(rule);
+                            }
+                            else if(refSubObjectType != null &&
+                                    refSubObjectType.equals(ruleSubObjectType)) {
+                                filteredParsingRules.add(rule);
+                            }
+                        }
+                        Collections.sort(filteredParsingRules, new ParsingRulesSubtypeFirstComparator());
+                        //List<ObjectData> data = new ArrayList<>();
+                        for (final ObjectTypeParsingRules rule : filteredParsingRules) {
+                            final ParseObjectsRet parsedRet = parseObjects(guid, this,
+                                    newRefPath, obj, rule);
+                            GUID parsedGUID = parsedRet.guidToObj.keySet().iterator().next();
+                            ObjectData.Builder objDataBuilder = ObjectData.getBuilder(parsedGUID,rule.getGlobalObjectType());
+                            Map<String, List<Object>> keywords = parsedRet.guidToObj.get(parsedGUID).getKeywords();
+                            Iterator<String> keys = keywords.keySet().iterator();
+                            while(keys.hasNext()) {
+                                String key = keys.next();
+                                String textValue;
+                                Object objValue = keywords.get(key);
+                                //////////////////////////////////////////////
+                                if (objValue instanceof List) {
+                                    @SuppressWarnings("unchecked")
+                                    final List<Object> objValue2 = (List<Object>) objValue;
+                                    textValue = objValue2.stream().map(Object::toString)
+                                            .collect(Collectors.joining(", "));
+                                } else {
+                                    textValue = String.valueOf(objValue);
+                                }
+                                //////////////////////////////////////////////
+                                objDataBuilder.withKeyProperty(key, textValue);
+                            }
+                            objDataBuilder.withNullableObjectName(obj.getName());
+                            objDataBuilder.withNullableCreator(obj.getCreator());
+                            objDataBuilder.withNullableMD5(obj.getCommitHash().orNull());
+                            data.add(objDataBuilder.build());
+                        }
+                    }
+                } catch(Exception ex) {
+                    ex.printStackTrace();
+                }
+
             }
+            return data;
+//            try {
+//                List<ObjectData> data2 = indexingStorage.getObjectsByIds(guids, pp);
+//                return data2;
+//            } catch (IOException e) {
+//                throw new RetriableIndexingException(ErrorType.OTHER, e.getMessage(), e);
+//            }
+                //////////////////////////////////////////////////////////////////
         }
         
         @Override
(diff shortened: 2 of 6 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,11 @@
 Search Service MKII release notes
```
MrCreosote, line 1: Changes could def use unit tests vs. just integration tests

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                                    obj.getName(),
+                                    obj.getCreator(),
+                                    obj.getCommitHash(),
+                                    parsedRet.guidToObj.get(parsedGUID).getKeywords());
+
+                            data.add(objData);
+                        }
+                    }
+                } catch(Exception ex) {
+                    ex.printStackTrace();
```
MrCreosote: This block looks extremely evil. Is it intentional?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
-            pp.objectKeys = true;
-            try {
-                return indexingStorage.getObjectsByIds(guids, pp);
-            } catch (IOException e) {
-                throw new RetriableIndexingException(ErrorType.OTHER, e.getMessage(), e);
+
+            final List<ObjectData> data = new ArrayList<>();
+
+            // get object from data source
+            for (GUID guid: guids) {
```
MrCreosote: This block looks pretty similar to the indexObjects() method IIRC. Is it possible to DRY it up? Or portions of it?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                try {
+                    final EventHandler handler = getEventHandler(guid);
+                    final Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
+                    while (refs.hasNext()) {
+                        final ResolvedReference ref = refs.next();
+
+                        // make a copy to avoid mutating the caller's path
+                        final LinkedList<GUID> newRefPath = new LinkedList<>();
+                        newRefPath.add(guid);
+                        final SourceData obj = handler.load(newRefPath, tempFile.toPath());
```
MrCreosote: Is reusing the same temp file an issue? Not sure.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
         
         private List<ObjectData> getObjectsByIds(final Set<GUID> guids)
```
MrCreosote: I believe this interface will need a change to allow passing in the full ref path so far. Otherwise one or more of the root guids may be deleted in which case the fetch will fail. The resolveRefs function made passing the full ref path unnecessary here since the data was fetched from ES, not the WS.

MrCreosote: It might be simpler to do the parsing in resolveRefs and cache the parsed data there, whether in memory or on disk.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                        }
+                    }
+                } catch(Exception ex) {
+                    ex.printStackTrace();
+                }
+            }
+            return data;
+        }
+
+        private List<ObjectTypeParsingRules> getFilteredSortedParsingRules(ResolvedReference ref) {
```
MrCreosote: Is this guaranteed to return only one rule? I *think* so but maybe I'm wrong.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                        final ResolvedReference ref = refs.next();
+
+                        // make a copy to avoid mutating the caller's path
+                        final LinkedList<GUID> newRefPath = new LinkedList<>();
+                        newRefPath.add(guid);
+                        final SourceData obj = handler.load(newRefPath, tempFile.toPath());
+
+                        final List<ObjectTypeParsingRules> filteredParsingRules = getFilteredSortedParsingRules(ref);
+
+                        for (final ObjectTypeParsingRules rule : filteredParsingRules) {
```
MrCreosote: This seems weird to me - it implies that the function takes a list of guids and then returns a list of object data where the guid index may not match with its corresponding object data index.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                try {
+                    FileUtil.getOrCreateSubDir(rootTempDir, guid.getStorageCode());
+                    tempFile = File.createTempFile("ws_srv_response_", ".json");
+                } catch (IOException ex) {
+                    throw new RetriableIndexingException(ErrorType.OTHER, ex.getMessage(), ex);
+                }
+
+                try {
+                    final EventHandler handler = getEventHandler(guid);
+                    final Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
```
MrCreosote: Why not just pull the RR from the cache? It would've been added in resolveRefs().

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -748,10 +749,6 @@ private void renameAllVersions(final GUID guid, final String newName)
                     final GUID guid = rr.getResolvedReference();
                     final boolean indexed = retrier.retryFunc(
```
MrCreosote: Looks like this line is unnecessary

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,15 +815,112 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                    throw new RetriableIndexingException(ErrorType.OTHER, ex.getMessage(), ex);
+                }
+
+                try {
+                    final EventHandler handler = getEventHandler(guid);
+                    final Iterator<ResolvedReference> refs = handler.resolveReferences(null, guids).iterator();
+                    while (refs.hasNext()) {
+                        final ResolvedReference ref = refs.next();
+
+                        // make a copy to avoid mutating the caller's path
```
MrCreosote: But you're not using a path here? The equivalent to the next 3 lines is:
`final SourceData obj = handler.load(Arrays.asList(guid), tempFile.toPath());`

RELEASE_NOTES.md
arfathpasha: I have full coverage for the change. I am not sure if it is worth the time since there are existing unit tests that test the base functionality.  If you want this I'll have to spend more time coding up these unit tests. I also feel we should have at least one integration test per type that we can use to test against the three or four environments we have.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Removed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I have changed the interface to include the object ref path. I don't really have a good understanding of the need for ref chaining. I have only a partial understanding at best. So please scrutinize this change in the commit "changed lookup provider interface"

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I tried but it is not quite the same. Some lines in the code flow are different.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I'm not seeing a cache of resolved references that I can use at this point in the code and I don't understand this whole concept of resolving references. I think a call to discuss this might help.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Added refpath but I don't know if this is right or wrong as I don't understand the resolved path concept fully. Please check.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Separation is always good.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I have fixed this. But it is now possible that an object will not get returned for a guid if the parsing rules cannot be found. I don't like the silent failure but it seems to be the underlying theme of the code for this case. I tried throwing an indexing exception but that exception has a protected constructor in a different package. It doesn't make sense to me. I should be able to generate an indexing exception from the indexer worker.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: You are right. It now returns one parsing rules obj or null if the parsing rules cannot be found.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: :O Fixed.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: Have you seen this: https://docs.google.com/presentation/d/1_SrBxGbWU53UNn2sM6KhR2fsbywjjxi2qn3mtloU81A/edit#slide=id.p

Explains why ref paths/chains are important.

RELEASE_NOTES.md
MrCreosote: > I am not sure if it is worth the time since there are existing unit tests that test the base functionality.

I don't think there are unit tests that test the recursion are there?

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: We should definitely hunt and kill as many silent failures as possible. They are godless ravening fiends.

IndexingException is the superclass for all non-retriable indexing exceptions - if the error isn't retriable, use one of the subclasses. If it's retriable, use RetriableIndexingException or a subclass.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: The cache is here: https://github.com/kbase/KBaseSearchEngine/blob/26f74aee7e733d0908e3cd4a2eb3336f253beb87/lib/src/kbasesearchengine/main/IndexerWorker.java#L713

MrCreosote: There's two reasons for resolving references. Let's say data A has a reference to data B in some data source. That reference may or may not be valid and may or may not be specified with a permanent address, and so the ref resolving method contacts to the data source to a) ensure the ref is valid and b) translate the ref to a permanent address.

The second reason is that, at least in the case of the workspace, if an object is deleted the *only* way to access it is via a ref path, even for admins. There are technical reasons for this that I won't go into here. So if Genome G depends on Assembly A and assembly A is deleted, a the indexing recursion needs to supply a ref path from G to A so A can be accessed, parsed, and any relevant fields flattened into the ES document for G.

If this still isn't clear let's have a slack call this afternoon.

RELEASE_NOTES.md
arfathpasha: Not specifically the recursion but there are tests in indexer worker that exercise the recursion. guidNotFoundError() for example.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I understand this now. I think my implementation resolves references now. 

I can think of ways to make the refpath problem opaque to workspace users, but that's another topic.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I have fixed this by throwing an exception. But it broke your guidNotFound test since the new exception gets thrown before the guidNotFound exception. I modified that test some to accommodate for this change.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I spent hours trying to use refResolvingCache from getObjectsByIds and found it doesn't work because the cache stores refpaths without subobjId info and getObjectByIds processes guids with subObjectIds. So when the cache lookup takes place, the resolved reference for a guid with subObjectId results in a reference to the main object and not the subObject. This eventually results in a guidNotFoundError because in KeywordParser,

```
final Map<GUID, SearchObjectType> guidToType = lookup.getTypesForGuids(objectRefPath, guids)
``` 
the guidToType contains the map for the main object and then fails on, 

```
if (!guidToType.containsKey(guid)) {
       throw new GUIDNotFoundException("GUID " + guid + " not found");
}
```
as the guid contains the subObjectId info.

MrCreosote: just resolve the main part of the GUID, not the subobject stuff.

RELEASE_NOTES.md
MrCreosote: I still think that units that if we're making large changes to IndexerWorker, the steps should be 
1) Add unit tests that cover the parts we're going to change
2) Make the changes and modify any unit tests that are invalid

That ensures the changes don't break anything else. Also, we should be adding missing unit tests as we go anyway.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                                unresolvedGUID.getSubObjectId());
+                resRef = new ResolvedReference(resRef.getReference(),
+                                               guid,
+                                               resRef.getType(),
+                                               resRef.getTimestamp());
+
+                // get object from data source
+                final File tempFile;
+                try {
+                    FileUtil.getOrCreateSubDir(rootTempDir, guid.getStorageCode());
```
MrCreosote: I'm confused about what's going on here. It looks like the subdir created on this line isn't used on the next line.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+         * @throws IndexingException
+         * @throws InterruptedException
+         * @throws RetriableIndexingException
+         */
+        private List<ObjectData> getObjectsByIds(final List<GUID> objectRefPath,
+                                                 final Set<GUID> unresolvedGUIDs)
+                throws IndexingException, InterruptedException, RetriableIndexingException {
+            final List<ObjectData> data = new ArrayList<>();
+
+            for(GUID unresolvedGUID: unresolvedGUIDs) {
```
MrCreosote: space after for, final GUID

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                final File tempFile;
+                try {
+                    FileUtil.getOrCreateSubDir(rootTempDir, guid.getStorageCode());
+                    tempFile = File.createTempFile("ws_srv_response_", ".json");
+                } catch (IOException ex) {
+                    throw new RetriableIndexingException(ErrorType.OTHER, ex.getMessage(), ex);
+                }
+
+                try {
+                    // make a copy to avoid mutating the caller's path
```
MrCreosote: This comment doesn't make sense AFAICT

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                if (refSubObjectType == null && ruleSubObjectType == null) {
+                    return rule;
+                }
+                else if (refSubObjectType != null &&
+                        refSubObjectType.equals(ruleSubObjectType)) {
+                    return rule;
+                }
+            }
+
+            return null;
```
MrCreosote: Why not just throw an exeption here?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                    tempFile = File.createTempFile("ws_srv_response_", ".json");
+                } catch (IOException ex) {
+                    throw new RetriableIndexingException(ErrorType.OTHER, ex.getMessage(), ex);
+                }
+
+                try {
+                    // make a copy to avoid mutating the caller's path
+                    final EventHandler handler = getEventHandler(guid);
+                    final SourceData obj = handler.load(
+                            Arrays.asList(resRef.getResolvedReference()), tempFile.toPath());
```
MrCreosote: This is where you'd want to use the reference path, as this is where the failure would occur if the target object is deleted and no reference path is provided.

A test case for this would be good.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+
+                    final ObjectTypeParsingRules rules = getObjectTypeParsingRules(resRef);
+
+                    if(rules == null) {
+                        throw new FatalIndexingException(ErrorType.OTHER,
```
MrCreosote: This will shut the worker down - is that the desired behavior? UnprocessableEventException is an option if not.

MrCreosote: If it's not already documented, should note that a spec is required for any dependent objects that need to be processed because they're referenced by a parent.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                                "Unable to find parsing rules for reference "+resRef);
+                    }
+
+                    final ParseObjectsRet parsedRet = parseObjects(guid, this,
+                            Arrays.asList(resRef.getResolvedReference()), obj, rules);
+
+                    final GUID parsedGUID = parsedRet.guidToObj.keySet().iterator().next();
```
MrCreosote: IIUC, this is taking the first entry in the results of the parse and only processing that. If I'm right, why are all the other possible entries ignored?

RELEASE_NOTES.md
```diff
@@ -1,6 +1,11 @@
...
 =================================
```
MrCreosote: I realize I need a primer on how the recursive indexing works.

1) Can the recursive indexing target a subobject for data extraction, or only the parent object?
2) Can ResolvedRefs ever contain GUIDs with subobject IDs? I don't think they're every produced like that from the event handler, but I don't think that behavior is documented.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: This is uncovered per the coverage report.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                final String textValue;
+                final Object objValue = keywords.get(key);
+
+                if (objValue instanceof List) {
+                    @SuppressWarnings("unchecked")
+                    final List<Object> objValue2 = (List<Object>) objValue;
+                    textValue = objValue2.stream().map(Object::toString)
+                            .collect(Collectors.joining(", "));
+                } else {
+                    textValue = String.valueOf(objValue);
```
MrCreosote, line 946: This is uncovered per the coverage report.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+        private ObjectData buildObjectDataFrom(GUID guid,
+                                               SearchObjectType globalObjectType,
+                                               String objectName,
+                                               String objectCreator,
+                                               Optional<String> md5,
+                                               Map<String, List<Object>> keywords) {
+
+            ObjectData.Builder objDataBuilder = ObjectData.getBuilder(guid,globalObjectType);
+            Iterator<String> keys = keywords.keySet().iterator();
+            while (keys.hasNext()) {
```
MrCreosote: why not just
```
for (final String key: keywords.keySet()) {
```

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -841,18 +801,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+
+
+        private ObjectData buildObjectDataFrom(GUID guid,
+                                               SearchObjectType globalObjectType,
+                                               String objectName,
+                                               String objectCreator,
+                                               Optional<String> md5,
+                                               Map<String, List<Object>> keywords) {
+
+            ObjectData.Builder objDataBuilder = ObjectData.getBuilder(guid,globalObjectType);
```
MrCreosote: final

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -875,7 +956,7 @@ public ObjectTypeParsingRules getTypeDescriptor(final SearchObjectType type)
             }
             if (guidsToLoad.size() > 0) {
                 final List<ObjectData> data =
-                        retrier.retryFunc(g -> getObjectsByIds(g), guidsToLoad, null);
+                        retrier.retryFunc(g -> getObjectsByIds(objectRefPath, g), guidsToLoad, null);
```
MrCreosote: This seems like a really expensive way to get the type of the object since it now fetches the object from the workspace just to get a type, which can be fetched much more cheaply.

arfathpasha:
Type fetching in MOPLookupProvider has been optimized.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Optimized type fetching in MOPLookupProvider

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Fixed. Now it is.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I didn't understand it either. Removed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I have cleaned up this code. Hopefully, I have it right now.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Using refpath now, but tests are still needed.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Changed to non-fatal exception

arfathpasha: added documentation

RELEASE_NOTES.md
arfathpasha: I need a primer every time I look at that bit of code. I think recursive indexing does target subobjects. I had to put in the subobject id into MOPProvider.getObjectsByIds() method to get it all to work and I forget now the exact reason for this. I'd need to look it up again :(

lib/src/kbasesearchengine/system/TypeFileStorage.java
```diff
@@ -294,4 +294,18 @@ public ObjectTypeParsingRules getObjectTypeParsingRules(final SearchObjectType t
         }
         return ret;
     }
+
+    @Override
+    public Set<SearchObjectType> getSearchObjectType(StorageObjectType type) {
```
MrCreosote, line 297: Seems like it'd make more sense to refactor this so the first part of the method above is in this method, and then the method above calls this method to get the SearchObjectTypes.

lib/src/kbasesearchengine/system/TypeFileStorage.java
```diff
@@ -294,4 +294,18 @@ public ObjectTypeParsingRules getObjectTypeParsingRules(final SearchObjectType t
...
+        Set<ObjectTypeParsingRules> rules = listObjectTypeParsingRules();
+        Set<SearchObjectType> seTypes = new HashSet<>();
+
+        for (final ObjectTypeParsingRules rule: rules) {
+            if ( type.equals(rule.getStorageObjectType())) {
```
MrCreosote: Isn't this guaranteed to be true?

RELEASE_NOTES.md
```diff
@@ -13,6 +13,11 @@ Search Service MKII release notes
 * In KBaseSearchEngine.spec, moved the flags addNarrativeInfo and addAccessGroupInfo
   from match filter parameter to the postprocessing parameter provided to the search API.
 
+0.2.0
```
MrCreosote: If 0.2.1 has already been released or is in preparation shouldn't this be in 0.2.1?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -513,10 +517,11 @@ private void indexObject(
          * 
          */
         long t1 = System.currentTimeMillis();
+        final File subDir;
```
MrCreosote: Looks like this isn't used outside the block, might as well move the declaration inside

lib/src/kbasesearchengine/parse/KeywordParser.java
```diff
@@ -268,7 +270,14 @@ private static Object transform(
             } catch (IllegalArgumentException e) {
                 throw new ObjectParseException(e.getMessage(), e);
             }
-            Set<GUID> guids = lookup.resolveRefs(objectRefPath, unresolvedGUIDs);
+            Set<ResolvedReference> resrefs = lookup.resolveRefs(objectRefPath, unresolvedGUIDs);
```
MrCreosote, line 273: I'm not understanding the need for this change - the resolved refs are never used other than to make the set of guids which were originally returned.

MrCreosote:
merge conflict here

RELEASE_NOTES.md
```diff
@@ -13,6 +13,11 @@ Search Service MKII release notes
 * In KBaseSearchEngine.spec, moved the flags addNarrativeInfo and addAccessGroupInfo
```
MrCreosote, line 23: Same old comment in that there should be unit tests for the part of the indexer worker that's getting changed

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -865,28 +950,45 @@ public ObjectTypeParsingRules getTypeDescriptor(final SearchObjectType type)
         }
         
         @Override
-        public Map<GUID, SearchObjectType> getTypesForGuids(Set<GUID> guids)
+        public Map<GUID, SearchObjectType> getTypesForGuids(final List<GUID> objectRefPath, final Set<GUID> unresolvedGUIDs)
```
MrCreosote: Based on reading the code, it seems like this method and resolveReferences are always called together and theoretically could be combined, simplifying the code. I leave that to you do decide if you want to do that.

MrCreosote: This part of the codebase is such a pain to understand any simplification is good imo...

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -865,28 +950,45 @@ public ObjectTypeParsingRules getTypeDescriptor(final SearchObjectType type)
...
                 throws InterruptedException, IndexingException {
             Map<GUID, SearchObjectType> ret = new LinkedHashMap<>();
-            Set<GUID> guidsToLoad = new LinkedHashSet<>();
-            for (GUID guid : guids) {
+            Set<ResolvedReference> guidsToLoad = new LinkedHashSet<>();
+            for (GUID unresolvedGUID : unresolvedGUIDs) {
+                ResolvedReference resRef =
```
MrCreosote: Why not just do this outside the loop and therefore do it once?

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,18 +808,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                ResolvedReference ref) throws ObjectParseException {
+
+            final List<ObjectTypeParsingRules> parsingRules = new ArrayList<>(
+                    typeStorage.listObjectTypeParsingRules(ref.getType()));
+
+            // filter by subObject type
+            final String refSubObjectType = ref.getResolvedReference().getSubObjectType();
+            for (final ObjectTypeParsingRules rule : parsingRules) {
+                final String ruleSubObjectType = rule.getSubObjectType().orNull();
+                if (refSubObjectType == null && ruleSubObjectType == null) {
```
MrCreosote, line 913: This assumes there's one non-subobject ObjectTypeParsingRule per source (e.g. workspace) type. Is that a correct assumption? I'm not sure.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -844,18 +808,139 @@ private void indexObjectWrapperFn(final List<Object> input)
...
+                                unresolvedGUID.getSubObjectType(),
+                                unresolvedGUID.getSubObjectId());
+                resRef = new ResolvedReference(resRef.getReference(),
+                                               guid,
+                                               resRef.getType(),
+                                               resRef.getTimestamp());
+
+                // get object from data source
+                final File tempFile;
+                final File subDir;
```
MrCreosote, line 850: Looks like this isn't used outside the block, might as well move the declaration inside

RELEASE_NOTES.md
MrCreosote, line 23: This code really needs a refactor / rewrite, it's too hard to grok. I can't tell if it's right or not because the input constraints are too loose.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: I'm still confused here. ParseObjectRet can contain many entries in guidToObj, but here all but the first one (which is arbitrary since it's a map) are ignored. That seems like a bug unless I'm missing something.

lib/src/kbasesearchengine/system/TypeFileStorage.java
arfathpasha: Done.

lib/src/kbasesearchengine/system/TypeFileStorage.java
arfathpasha: This code was removed as part of the previous refactoring.

RELEASE_NOTES.md
arfathpasha: Moved note to 0.2.1.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: lookup.resolveRefs returns a Set<ResolvedReference> which is then being converted to a Set<GUID>. This change was made to use cached resolvedRefs. 

https://github.com/arfathpasha/KBaseSearchEngine/commit/fd71a20e451f9b9fc0bafb10191f384f9616b04e#diff-14fdf7569af870a20d0ae2cc960997c6

RELEASE_NOTES.md
arfathpasha: Not sure I follow. The comment references an unrelated release note.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I see your point, but each time I try to clean up this code I feel like I am entering an abyss. I think it should be done as a separate task so it doesn't hold up this one.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: That would be an optimization that is based on the current use of the method. If the use changes, for instance by sending in a mixed bag of guids, this optimization will become a breaking change. For this reason, I would rather leave it as is.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I think this is the case because sub-object types key off the parent type. So even if two parents had identical sub-objects, the sub-obj types would have to be defined separately with each one keying off of its respective parent.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: sub-object type specs i.e.

RELEASE_NOTES.md
arfathpasha: Same as above. The comment references an unrelated release note, so I'm not following.

arfathpasha:
The ParseObjectRet issue has been fixed with an iterator that loops over the entire keySet.

RELEASE_NOTES.md
MrCreosote: It's just a general comment.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: > I think this is the case because sub-object types key off the parent type. 

Is that true? I'm looking at the AssemblyContig spec and it doesn't look to me like there's any references to the Assembly type at all. IFAIK the Assembly spec could be deleted and everything would work fine. Am I incorrect here?

RELEASE_NOTES.md
MrCreosote: Checking the master branch release notes, 0.2.1 was already released (although no releases other than 0.1.0 are tagged in git), so this probably needs to go in 0.2.2.

lib/src/kbasesearchengine/parse/KeywordParser.java
MrCreosote: Sorry, I'm still confused here. resolveRefs() was changed to return Set\<ResolvedReference> rather than Set\<GUID>. But this is the only place in the codebase where the method is called, and it immediately changes the Set\<ResolvedReference> back to Set\<GUID> and never uses the Set\<ResolvedReference> at all. 

It seems to me that there's no reason for this change, and I don't understand how caching the ResolvedReferences has anything to do with it - that happens regardless of the return type right?

RELEASE_NOTES.md
MrCreosote: I just stuck this comment at the top of the PR diff since to have a comment as part of a review it needs to be anchored somewhere, AFAICT. The context isn't relevant.

Anyway, the comment refers to this:

> I still think that units that if we're making large changes to IndexerWorker, the steps should be
>
> 1. Add unit tests that cover the parts we're going to change
> 2. Make the changes and modify any unit tests that are invalid
>
> That ensures the changes don't break anything else. Also, we should be adding missing unit tests as we go anyway.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: > each time I try to clean up this code I feel like I am entering an abyss

That's pretty much how I feel trying to review it. It's really hard to tell if the changes are correct or not because I don't really grok the original code.

lib/src/kbasesearchengine/parse/KeywordParser.java
MrCreosote: Maybe I should clarify that 'this change' = the return type of the method, not necessarily changes to the method implementation itself.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: Not sure I'm following. I mean something like this:

```java
final Set<RR> rrs = resolveRefs(objectRefPath, unresolvedGUIDs);
for (final RR resRef: rrs) {
    final GUID guid = resRef.getResolvedReference();
...
```

That's exactly the same as the current implemenation except there's 1 call to resolveRefs vs. N.

RELEASE_NOTES.md
arfathpasha: I couldn't agree me. I have never struggled this much to fix code.

RELEASE_NOTES.md
arfathpasha: Done.

lib/src/kbasesearchengine/parse/KeywordParser.java
arfathpasha: resolveRefs() is used in three places - 

https://github.com/arfathpasha/KBaseSearchEngine/blob/dev-recursive-indexing/lib/src/kbasesearchengine/main/IndexerWorker.java#L831

https://github.com/arfathpasha/KBaseSearchEngine/blob/dev-recursive-indexing/lib/src/kbasesearchengine/main/IndexerWorker.java#L964

https://github.com/arfathpasha/KBaseSearchEngine/blob/dev-recursive-indexing/lib/src/kbasesearchengine/parse/KeywordParser.java#L273

I have not analyzed to see if they all cascade down to a single use since there is caching involved and this should compensate for any inefficiency. I think my overall goal here was to make sure I resolve a guid before I use it simply as a safety measure.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: I guess keying off is not the right terminology. The line below states that AssemblyContig is an inner-sub-type of the parent type Assembly right? This is what I meant.   

storage-object-type: KBaseGenomeAnnotations.Assembly

MrCreosote: Both the search types Assembly and AssemblyContig relate to the same storage type (e.g. workspace type), KBaseGenomeAnnotations.Assembly, but AFAICT are otherwise completely independent. You could have search installations that index both, only Assembly search types, or only AssemblyContig search types. AFAIK there could also be other non-subobject types related to the storage type.

To the best of my ability to tell, that means that when the code assumes there's exactly one top-level / non-subobject type per storage type, as I think it does in this case, that's a bug, because there could be 0, 2, or any other number.

lib/src/kbasesearchengine/parse/KeywordParser.java
MrCreosote: Doh, when I searched the diff IndexerWorker hadn't been loaded by github so those uses didn't show up.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: OK, I spent an hour trying to figure out why I did it this way after spending an hour figuring the same thing out the last time I did it this way. The reason is because the subObjectType and subObjectId gets lost in the resolveReference process and the only way I can get it is by iterating through the unresolved guids and then getting this info from the unresolved guid in each iteration. 

The reason why this info is lost is because the cache used in the line below keys on a ref path that includes only parent ref info (1/2/1;1/3/1 for example). So by the time the sub-objects get resolved, they get the parent info and lose all sub-object info. 

https://github.com/arfathpasha/KBaseSearchEngine/blob/master/lib/src/kbasesearchengine/main/IndexerWorker.java#L738

MrCreosote: I see. In that case, IIUC, you could do

```java
final Set<RR> rrs = resolveRefs(objectRefPath, unresolvedGUIDs);
final Map<GUID, RR> rrmap = rrs.stream().toMap(rr -> rr.getReference(), rr);
for (final GUID unresolvedGUID: unresolvedGUIDS) {
    final RR resRef = rrmap.get(unresolvedGUID.parentGUID());
    ...
```

Not sure if I have all the syntax and method calls right but you get the idea.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: By source type, do you mean storage type?

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Fixed this with an iterator.

lib/src/kbasesearchengine/main/IndexerWorker.java
arfathpasha: Done.

lib/src/kbasesearchengine/main/IndexerWorker.java
MrCreosote: Yes, edited to make the names consistent.

arfathpasha: I think I see what you mean now. I was thinking in a different direction (a sub-object having two parent types). But I think what you mean is that there can be many specs defined for a single storage object type. So in this case, we would need to populate more than one index for each guid with a storage object type that maps to multiple search object types (sub-object types seem to add a layer of complexity and will also have to be handled in a similar manner). Does this rephrasing of the problem/solution sound right? I want to make sure before I make the change since the change cascades into other parts.

MrCreosote: > (a sub-object having two parent types)

I don't think I understand what you mean by this.

Your rephrasing sounds right, but just in case, here's another rephrasing:

For each storage type, there may be X search type specs and Y search sub-type specs, where X and Y are integers from 0 to N. There is no relationship between the various specs other than they happen to be for the same storage type.

The current code handles this situation correctly because it just pulls the list of search types for the storage type from the type storage system and processes them one by one. This new code, if I understand it correctly, does not do that and makes assumptions about the values of X and Y.

arfathpasha: After analyzing it more, I think the app is still working as we expect it to. The reason for this is that all X types refer back to the same storage type. The same is true for the Y sub-types. So the code in the snippet above will get back that storage type. 

As a sanity check I ran this PR version against CI with two genome.yaml files (with search types Genome and GenomeNew) referring back to the same storage type and added the mapping shown below and it generated the corresponding genome_2 and genomeNew_2 indexes. This I think is also the use case that Shane described in our reindexing discussion yesterday. 

```
storage-type: WS
types:
    KBaseGenomes.Genome:
        default:
           -
               type: Genome
               version: 2
           -
               type: GenomeFeature
               version: 2
           -
               type: GenomeNew
               version: 2
```

MrCreosote: I don't think it can be working right. I think the bug will occur if 1) there's 2 or more specifications for a storage type and 2) a recursion encounters that type.

What the code is doing is, given a storage type and a guid (which in the case of the bug, does not have a sub-object type):

1) Get all the object type parsing rules for the storage type, of which there are N
2) Loops though the rules until it finds one without a subobject type
3) Returns that rule.

That means that any other rules without sub-object types are ignored.

Actually, on reading further, it's the same problem for sub-object types - it finds the first rule that matches the sub-object type  and returns it, ignoring any other rules.

arfathpasha: The reason why it looks like it is working correctly is because the objects get indexed in the code below. 

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/main/IndexerWorker.java#L538

The code snippet above is used in two places.

The code is being used to perform a lookup transform here -

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/parse/KeywordParser.java#L313

The code is being used to find the search object type here - 

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/parse/KeywordParser.java#L294

In both cases, the code is being used for performing transforms. This code will blow up only when a transform with a reference to a storage type that maps to two or more search types is made. This code AFAICT is not a regression and points to inherent bugs in the application. I am now working on fixing this.

MrCreosote:
Sorry, I got this confused with a different PR in the standup, I think - is this one ready for review?

arfathpasha:
Yes this is ready for review. I have fixed the bug in the lookup code that is being used by the keyword parser and I have added a couple of tests.

MrCreosote:
Can you describe the bug you say was already in the code base but that you fixed in this PR? I'm not totally sure I understand it.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -832,7 +797,7 @@ private void indexObjectWrapperFn(final List<Object> input)
             }
             if (guidsToLoad.size() > 0) {
                 final List<ObjectData> objList =
-                        retrier.retryFunc(g -> getObjectsByIds(g), guidsToLoad, null);
+                        retrier.retryFunc(g -> getObjectsByIds(objectRefPath, g), guidsToLoad, null);
```
MrCreosote, line 800: I think there's a bug here - IIUC, getObjectsByIds will return a list of ObjectData, and for each GUID (including sub-object GUIDS) there will be 1 ObjectData per search type. 

But then everything's put into a map of guid -> object data, so all but one instance of the object data will be lost.

arfathpasha: Yes this is one of two bugs I found. The other one is fixed and both are detailed below.

arfathpasha:
Here are the details about the bug - 

If we have multiple type specs for a given storage type, then we expect the indexer to pull the list of search types for the storage type from the type storage system and processes them one by one. This happens correctly in the for-loop in this code snippet. 

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/main/IndexerWorker.java#L538

But, when the type specs have transforms listed in them, the lookup transform code was basically pulling the transform info from one spec's ObjectTypeParsingRules and using those rules for indexing of all search types. The fix now makes the guid lookup transform code look at all the specs and apply each spec's transform rules separately for each of the search types being indexed. i.e. this fix is specific to the guid transforms that use the lookup provider code. 

So the line of code in the link 

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/parse/KeywordParser.java#L294

is modified from 

```
final Map<GUID, SearchObjectType> guidToType = lookup.getTypesForGuids(guids);
```
to 
```
final Map<GUID, List<SearchObjectType>> guidToType = lookup.getTypesForGuids(objectRefPath, guids);
```
As you pointed out, the same problem exists for lookup transforms in line 

https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/parse/KeywordParser.java#L313

I could have sworn I fixed this too, but I guess I didn't!

PR #241, opened 2018-04-30
SCT-653  Added _id field to indexer

arfathpasha:
- used RFC 4122 compliant UUID for ES _id. This can be produced with any programming language as libs are widely available. 
- ES _id field is always going to be 16 bytes long
- No information is lost as GUID can be indexed sperately

Additional PRs are required for, 
1. Refactoring ElasticSearchIndexingStorage.java to use the _id for lookups
2. Cleaning up GUID.java


MrCreosote:
Just based on reading the description, I'm confused by this. The point of the id swap was so we could do 
`PUT /[index]/[type]/[fn(GUID)]`.

If we're just swapping an id randomly generated by ES to an id randomly generated by a UUID library I don't understand the benefit.

MrCreosote:
Ok, that makes sense. My next question is around the readability of the ID. IIRC the maximum id length is 512 bytes which is probably fine for all the GUIDs we'll encounter (probably worst case is around 21 for the UPA + 4 for the storage code + 20 for the subobject id space + 100 for the subobject id + 3 for separators = 148 bytes). As such, it'd be great when working with kibana or curl to be able to GET and PUT with the GUID rather than having to compute a UUID.

Hence I was thinking of a fn along the lines of 
```
if (guid.length() < 512) {
    return guid.toString();
else {
    return guid.getParentGUID + ":" + UUID.uuid3(guid.getSubobjectNameAndID());
}
```

That'd allow us to use the standard guid in GETs and PUT in all but the most extreme cases (which I doubt we'll ever see), and even then, would make the GETs and PUTs slightly more readable.

arfathpasha:
I thought about this and my initial reaction was to KISS (keep it simple stupid) because the guid already exists in the indexes and can be easily searched for with,
```
GET INDEX/_search
{
  "query": {
    "match": {
      "FIELD": "TEXT"
    }
  }
}
```
With every conditional we add, we add complexity. In this case, we will have two ES id types. Not nice because it's yet another thing to keep in mind, albeit one will rarely be used. I will need to change the string representation of the guid to be something like ws-1-1-1-subname-subid so it can be used in a url without ugly escape chars. This change has cascading effects in the code, like messing with a parser that relies heavily on the current syntax. I'm thinking we should err on the side of keeping the code KISS. Lots to benefit from that per Roy's email about software minimalism.

arfathpasha:
Do you want to go with this format then?
Current:
{storageCode}:[{AccessGroupID}/]{accessGroupObjectId}[/{version}][:{subObjectType}/{subObjectId}]
New:
{storageCode}-[{AccessGroupID}.]{accessGroupObjectId}[.{version}][-{subObjectType}.{subObjectId}]

MrCreosote:
I'm ambivalent re format as long as it's easy to translate a 'standard' GUID to the url format and vice versa.

arfathpasha:
This is opening a can of worms as I suspected. 

1. Need to replace URI special characters with ascii characters
2. Need to escape delimiters in all the tokens. 
3. 1&2 can make the guid non-unique, so now I need to compensate for that. 

In the end, the guid may not look anything like the data source guid and this will make it hard to match the source and target objects up. Besides it is pretty difficult to type in these guids as ids because they can be fairly long strings and they have to be typed up in their entirety for an exact match. Copy-paste from a data source representation may not work because of all the transformations I listed above. Whereas the key-value search can use wildcards. I still strongly feel we should keep this simple.

arfathpasha:
AccessGroupObjectId will also need to be url encoded since (at least in theory) it could come from a data source outside of kbase as I understand. Curl GETs and PUTs will also have to url encode these parts, which is a pain. Partial treatment of segments of the id itself doesn't feel elegant.

MrCreosote:
> AccessGroupObjectId will also need to be url encoded since (at least in theory) it could come from a data source outside of kbase as I understand

I don't think it'd be outside of kbase, but rather from kbase systems other than the workspace. You're right in that it might not be an integer. For that one we should also add some reasonable constraints (maybe `[A-Za-z0-9]+` to start with).

I think the algorithm for this would be 1) make any necessary substitutions (like `/` -> `_` or whatever) 2) urlencode the whole thing, which should leave the storage code, UPA (less object id possibly), and subobject type untouched.

MrCreosote:
Note that a lot of applications will do the url encoding automatically. Not sure if curl and elastic search will do it, but browsers will.

arfathpasha:
```
curl -i -XGET http://localhost:9200/movies/movie/1$1?pretty
behaves like,
curl -i -XGET http://localhost:9200/movies/movie/1?pretty
Whereas
curl -i -XGET http://localhost:9200/movies/movie/1%241?pretty
behaves like,
curl -i -XGET http://localhost:9200/movies/movie/1$1?pretty
```
So there is no implicit url encoding of path segment in curl and there is no other explicit provision either. --data-urlencode urll encodes only for the query string portion and not for path segments. 

Kibana has implicit url encoding of path segments and query strings.

arfathpasha:
Do you still want me to add url-encoding? It doesn't quite work well with curl and I am concerned it could cause problems in other places too since it is one more thing that we now have to remember to do in the code base and outside of the code base - url-encode the id before using it.

MrCreosote:
I'm ambivalent about the solution to the problem, but I think the problem statement is (@scanon this is my understanding from our conversations, but please correct me if I'm wrong):

1) We wish to be able to GET and PUT to elasticsearch using GUIDs.
2) There are some cases where the GUIDs may cause problems:
  a) The GUID exceeds the ES 512 byte limit.
  b) The GUID contains characters that cause problems with URLS, particularly `/`.
3) Other than `/` we expect these cases to be extremely rare - probably almost nonexistant.

Since we expect the GUID to be acceptable nearly all of the time, it makes things a lot simpler for the user if they can GET/PUT with the GUID (possibly switching `/` for `_` or something) we should allow that, and document a procedure for dealing with the rare cases of nasty GUIDs. 

All that being said, if URL encoding is the solution, I don't think we'd actually want to url encode the GUID stored in elastic as the `_id` - the url encoding would have to happen in the application doing the GET and PUTs. Test to be sure of course.

arfathpasha:
@scanon any thoughts on this? Can we merge this? I am happy to make any changes you guys want but I wanted to state my concerns first.

lib/src/kbasesearchengine/common/GUID.java
```diff
@@ -46,6 +60,16 @@ public GUID(final GUID parentGUID, final String subObjectType, final String subO
         version = parentGUID.getVersion();
         this.subObjectType = subObjectType;
         this.subObjectId = subObjectID;
+
+        validate();
+    }
+
+    private void validate() {
+        final int guidBytes = toString().getBytes(StandardCharsets.UTF_8).length;
```
MrCreosote: The other option for really long GUIDs is to hash/uuidify the subobject ID (which is the only part that can be really long, most likely). 

I'm ok with throwing an exception as well. 

@scanon do you have a preference? If not let's go with the exception and try reindexing CI and see if we have any problems. Need to make sure we're indexing subobjects.

lib/src/kbasesearchengine/common/GUID.java
```diff
@@ -128,21 +152,54 @@ public GUID(String textGUID) {
...
+
+        String encodedGUID = toString("___", "_");
+
+        try {
+
+             encodedGUID =  URLEncoder.encode(encodedGUID, "UTF-8");
+
+        } catch(UnsupportedEncodingException ex ) {
+            // should never occur since the encoding is set to UTF-8 which is supported
+            throw new IOException(ex.getMessage());
```
MrCreosote: I'd throw a RuntimeException in that case. No need to make the user deal with an exception they'll never see.

lib/src/kbasesearchengine/common/GUID.java
```diff
@@ -128,21 +152,54 @@ public GUID(String textGUID) {
...
+     *
+     *   {storageCode}___[{AccessGroupID}_]{accessGroupObjectId}[_{version}][___{subObjectType}_{subObjectId}],
+     *
+     *  where special characters in the individual tokens will get URL-encoded.
+     *
+     * @return URL-encoded version of this GUID
+     */
+    public String getURLEncoded() throws IOException {
+
+        String encodedGUID = toString("___", "_");
```
MrCreosote: What's the reasoning for swapping ":" for "___"?

MrCreosote: If we're using "_" to separate the reference part, does that mean we need to make it an illegal character for object IDs?
Same question for subobject types.

lib/src/kbasesearchengine/common/GUID.java
```diff
@@ -128,21 +152,54 @@ public GUID(String textGUID) {
...
+
+        try {
+
+             encodedGUID =  URLEncoder.encode(encodedGUID, "UTF-8");
```
MrCreosote: I'm not clear on why we need to worry about urlencoding on the server side - that's a client side issue, right?

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: Mentioned this to @scanon and he's fine with an exception, but the exception should be specific so that we can search the DB for the exception type to see how many of these we get.

arfathpasha: Done.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: ":" encodes to "%3A" and makes the guid look ugly.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: KBSearch is a client to ElasticSearch. We use the ES REST API to interact with ES.

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: But why do you need to encode it at all? It should work fine without encoding even with curl:
```
crusherofheads@icrushdeheads:~$ curl "http://homology.kbase.us/namespace/KBase:Refseq" | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   248  100   248    0     0   1578      0 --:--:-- --:--:-- --:--:--  1579
{
    "error": {
        "appcode": 30001,
        "apperror": "Illegal input parameter",
        "callid": "4105873853039241",
        "httpcode": 400,
        "httpstatus": "Bad Request",
        "message": "30001 Illegal input parameter: Illegal character in namespace id KBase:Refseq: :",
        "time": 1527730889932
    }
}
```

MrCreosote: If I understand what you're saying, the ID transformation is one way?

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: Anything that's a RuntimeException is not going to wind up in the recorded in the db, it's just going to kill the thread.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: Yep

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: yeah, that didn't make sense. What I'm trying to say is that I don't think the stored ID needs to be urlencoded - the client (in this case the ES REST client) will handle the urlencoding when it sends the request, and the ES server will unencode it before presenting it to the business logic. That's the way I'd expect it to work anyway.

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: Fair enough then.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: I turned it into a runtime exception out of laziness. This code is used in over 500 places. I can make the updates to link the checked exceptions if you went. It's going to be a lot of updates and I'm not sure it's worth it for a rare case

MrCreosote: The point of the special exception was to be able to check the DB and see if we have any problems with long IDs when we index CI, so switching the exception to a Runtime is kind of pointless. Maybe check the length issue outside the GUID class prior to the load in a part of the code that already throws IndexingExceptions?

arfathpasha: That would defeat the guid validation. I'll give the checked exception linking a crack and see if I can wing it without it getting out of hand

MrCreosote: Or just catch and wrap the unchecked exception at the appropriate place and document the unchecked exception.

arfathpasha: Same thing right? I'd have to cover for over 500 possible places where this exception would need to be caught.

MrCreosote: I don't think so, just catching it here would work:
https://github.com/kbase/KBaseSearchEngine/blob/develop/lib/src/kbasesearchengine/main/IndexerWorker.java#L650
All the sub object IDs are created in there I think.

arfathpasha: I gave it a try. It is getting really ugly. Here are the files I have modified and I am still not done. I feel the unchecked exception is the best bet here. It's highly unlikely we are going to run into a guid that is 500 chars long - a guid that is 5 lines long. 

```
	modified:   lib/src/kbasesearchengine/common/GUID.java
	modified:   lib/src/kbasesearchengine/common/GUIDTooLongException.java
	modified:   lib/src/kbasesearchengine/events/StatusEvent.java
	modified:   lib/src/kbasesearchengine/events/exceptions/IndexingException.java
	modified:   lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
	modified:   lib/src/kbasesearchengine/main/IndexerCoordinator.java
	modified:   lib/src/kbasesearchengine/main/IndexerWorker.java
	modified:   lib/src/kbasesearchengine/parse/KeywordParser.java
	modified:   lib/src/kbasesearchengine/parse/ObjectParser.java
	modified:   lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
	modified:   lib/src/kbasesearchengine/search/ObjectData.java
	modified:   test/src/kbasesearchengine/test/common/GUIDTest.java
	modified:   test/src/kbasesearchengine/test/events/StatusEventTest.java
	modified:   test/src/kbasesearchengine/test/events/handler/ResolvedReferenceTest.java
	modified:   test/src/kbasesearchengine/test/parse/KeyWordParserTest.java

```

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: In your example the output states that the request was bad.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: I think our messages crossed. I have added a check for the unchecked exception as you suggested.

lib/src/kbasesearchengine/common/GUID.java
MrCreosote: But that's the business layer talking, not the http layer. The business layer got a colon, not the encoding for a colon.

arfathpasha: This is what I think is happening. ":" is a valid but reserved char in http. Since it is valid, http will not complain. But once this info is passed on to the receiving application, the info make no sense to the application because "KBase:Refseq" is probably being broken up into host segment and port by the http layer and then being supplied to the app layer. So the app layer complains as it doesn't understand the request.

MrCreosote: No, it's received as a single string.

arfathpasha: Then I guess it's because the app doesn't expect a ":" in a path segment. Not part of its API spec.

MrCreosote: It's because a colon is an illegal character in a namespace ID.

arfathpasha: Makes sense. I was wrong, the port ":" comes much before the path segments.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: Done.

lib/src/kbasesearchengine/common/GUID.java
arfathpasha: Since ES API doesn't have a problem with ":" in path segment I have switched "___" delim with ":". 
Here's what the _id looks like now - 
WS:1_1_1:contig_NZ_MCBT01000008

PR #243, opened 2018-05-07
SCT-755 Tested instructions for performing the 3 reindex cases

arfathpasha:
Docs contain some TODOs which will be addressed in future PRs.


docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+.. code-block:: text
+
+ alias -> [indices]
+ genome -> [genome_1, genome_2 ... genome_n]
+
+When reindexing is necessary, the index is reindexed to a new index with a new search type version. The first reindexing operation performed on genome_1 will result in a new index genome_1_r1 that replaces genome_1 in the genome alias. n reindexing operations performed on genome_1 will result in a new index genome_1_rn that replaces index genome_1_rn-1 in the genome alias.
+
+Reindexing Process
+-------------------
+For the sake of simplicity and for the reasons described below. a single process has been defined for all reindexing cases -
```
MrCreosote: below, a

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+
+.. code-block:: bash
+
+    GET kbase.genome_1/_mapping
+
+5b. Copy-paste the mapping from the current index into the body section of the PUT command below and make the necessary field change (preferably one change per complete reindexing operation).
+
+It is a good practice to make the mapping strict ("dynamic": "strict") for each type (data and access) in the index. Strict mappings prevent the mapping from being modified dynamically during ingest time.
+
+Update the settings section below the mapping. The number of shards and replicas must be decided on based on your capacity planning rules. It is costly to change the number of shards, so choose wisely! Make sure not to exceed 600 shards for any node in the system. Increase number of replicas to improve availability.
```
MrCreosote: Maybe note that:

* increase write speed = more shards
* increase read speed = more replicas
* max shard size = ~50GB

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+
+    GET /kbase.genome_1/_stats/docs,store
+
+5a. If mapping needs to be changed (for cases b, c, d above), get current index mapping.
+
+.. code-block:: bash
+
+    GET kbase.genome_1/_mapping
+
+5b. Copy-paste the mapping from the current index into the body section of the PUT command below and make the necessary field change (preferably one change per complete reindexing operation).
```
MrCreosote, line 184: Seems like a command for creating an index based on  a search type would reduce potential errors a lot.

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+      },
+      "settings": {
+        "index": {
+          "number_of_shards": "5",
+          "number_of_replicas": "1"
+        }
+      }
+    }
+
+5c. If the mapping does not require any change but the document's meta data needs to be changed, use the `Painless <https://www.elastic.co/guide/en/elasticsearch/reference/5.4/modules-scripting-painless-syntax.html>`_ script to modify metadata. Setting version_type to external will cause Elasticsearch to preserve the version from the source, create any documents that are missing, and update any documents that have an older version in the destination index than they do in the source index.
```
MrCreosote: Define meta data in this context, not clear.

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+
+10. List all available indexes for the genome alias and all available genome indexes to ensure consistency across the alias map. Verify that all genome indexes that are present (except for backups) are referenced by the alias. Also verify that the alias does not contain an index reference for which no index exists.
+
+.. code-block:: bash
+
+    GET /_cat/aliases/kbase.genome
+
+    GET /_cat/indices/kbase.genome_*
+
+11. If the change involved in the reindexing operation also requires a corresponding search type spec change (located in resources/types/genome.yml for example), then this change must be applied.
```
MrCreosote: applied to what? the spec?

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+
+12. Change mapping version from "1" to "_r1" in the resources/types/genome.yml search type spec and add a comment (for future reference) that describes the change that took place in the r1 reindexing operation.
```
MrCreosote: Shouldn't we abstract away this detail from the search specs? From my understanding of the discussion with @scanon, the plan was to have aliases for each index so we could point `genome_1`, say, at the current index and then the specs don't need to be aware of the reindexing count, which makes things a lot simpler for the spec authors.

docsource/ops.rst
```diff
@@ -1,6 +1,228 @@
...
+
+13. Restart trickle updates from the current time noted in step 1.
```
MrCreosote: Given these instructions I don't understand how reindexing would work when data needs to be fetched from the data source (say a new field is added to the spec).

resources/typemappings/GenomeAndAssembly.yaml.example
```diff
@@ -1,23 +1,28 @@
-# This is an example file that demonstrates how to specify storage-to-search type mappings beyond
+# This is an example file that demonstrates how to specify "storage-to-search" type mappings beyond
 # those embedded in the type transformation files (in the storage-type and storage-object-type
-# parameters).
+# parameters). Here, "storage" implies the data source storage system (mongodb for example) and "search"
```
MrCreosote: Not sure if mongodb is the best example here - I doubt any data source will expose mongo directly. The workspace is probably the best example we have right now.

MrCreosote (changes requested):

docsource/ops.rst
arfathpasha: Fixed.

docsource/ops.rst
arfathpasha: Done.

docsource/ops.rst
arfathpasha: I agree. I have made a note of this for a future PR.

docsource/ops.rst
arfathpasha: Added in brackets "field values".

docsource/ops.rst
arfathpasha: Yes. Added "...then this change must be applied to the spec as well.". Your idea about creating the index from spec will take care of this step too. I have left this for a future PR.

docsource/ops.rst
arfathpasha: The specs are the blue prints to the index mappings. If the mappings are changed, it will be useful for the spec authors to know about these changes so they don't make other changes that violate the reindexing changes. Also, we need a place to maintain a log of the reindexing changes. Since the specs are the blue prints, I felt this would be the best place to document reindexing changes. In addition, once we have a CLI to generate a new index from the spec, the reindex changes will be documented in the spec anyway.

docsource/ops.rst
arfathpasha: a) Change the type mapping, b) change the spec, generate new events for some time range in which the new field was added and run trickle updates (updates that take place as events trickle in). So far, I have defined a manual process with the steps a and b. Once we have a CLI for creating an index from a spec, it will be slightly more automated (with b being done before a) but the general idea still remains the same. This also assumes that we modify Search to overwrite existing data.

resources/typemappings/GenomeAndAssembly.yaml.example
arfathpasha: Fixed.

docsource/ops.rst
MrCreosote: Are you talking about the `_X` fields or the `key.X` fields? Or both?

docsource/ops.rst
MrCreosote: > generate new events for some time range in which the new field was added

I'm not seeing this in the instructions unless I'm missing something? It's always a reindex from the existing ES data AFAICT.

docsource/ops.rst
MrCreosote: How would this work? Right now the spec versions are keyed off their position in a list. Now a spec version will also have a reindexing version, right? So a spec maps to a specific index which consists of a type, a type version, and a reindexing version. Would the reindexing version be embedded in the spec? What if we reindex a different number of times in different environments?

docsource/ops.rst
arfathpasha: Ah, good point. Clarified as follows: 

 If the mapping does not require any change but the documents' field values (not including meta-data fields whose key names start with underscore) need to be changed, use the `Painless <https://www.elastic.co/guide/en/elasticsearch/reference/5.4/modules-scripting-painless-syntax.html>`_ script to modify metadata.

docsource/ops.rst
arfathpasha: Great questions. Rather than keying off of the implicit position in a list, we can make the spec versions explicit. This also has the effect of making the specs a lot more user friendly for the spec authors. So the spec could look like - 

```
storage-type: WS
storage-object-type: KBaseGenomes.Genome
versions:
    -
        version: 1
        indexing-rules:
            -
                path: id
                keyword-type: keyword
     . . .
    -
        # a comment about what reindexing operation r1 was all about
        version: 1_r1
        indexing-rules:
            -
                path: id
                keyword-type: keyword
    . . .
```
Here, the reindexing version is represented as a minor version number of the spec type version. The reason for this is because I am looking at reindexing operations as hot-fixes to problems that will ultimately need to be addressed by the type specs and the indexer. For example, let's say we decide to reindex to change a field value. This reindex operation can be done fairly quickly in prod and is akin to a hot-fix as this change can potentially be made available to users almost immediately. But ultimately, the same field value change will need to be made in the type spec as well with a lookup transform in order to keep the index and spec in synch. The latter will take relatively more time (hopefully not a whole lot longer if we have a well defined workflow and no code changes/maintenance to take care of for the change) as it has to go through a slower workflow but will be required in order to continue making the trickle updates to the new reindexed index via the indexer. In the example above, the synching update to the type spec version 1 can be represented as version 1_r1. 

So the reindexing workflow I am proposing is as follows: 
1. stop trickle updates
2. reindex (not recommended, but potentially can go live right here)
3. update type spec
4. start trickle updates on old and new index (new index through version 1_r1)
5. if all looks good, update the alias
6. at some point in the future consolidate versions 1_r1, 1_r2, ..., 1_rn into a new version 2. 

This way every move we make on the index is documented, synched up with the formal specs and executed through a well defined workflow.

Regarding different envs, we need to maintain separete specs for the different envs. This is why I suggested the following dir structure. 
```
resources
   |_ ci
       |_ typemappings
       |_ types
   |_ appdev
       |_ typemappings
       |_ types
...
```

I think it will be best to maintain this structure in KBaseSearchEngine itself as the specs are pretty integral with the application and after thinking about it for a while, I tend to disagree that the specs can be viewed as data that needs to be separated from the application. I view them as configurations for the application for the various envs. In fact, I think we should change the resources dir name to config. Since we have a finite set of envs, we can maintain them easily in github alongside the application and solve a lot of process and communication related problems this way.

docsource/ops.rst
arfathpasha: Ah, good point. Added.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+
+    The commands below can be copy-pasted into Kibana and executed against the index. The corresponding curl commands can be obtained from Kibana by clicking on the little wrench icon that appears next to the pasted command.
+
+.. note::
+
+    If any of the steps below fail, don't proceed until the issue is resolved by referring to the ElasticSearch documentation.
+
+1. Note current time
+
+2. Stop any workers that are performing trickle updates on the index that needs to be reindexed.
```
kkellerlbl: This might be nontrivial.  The current production workers aren't typically designated to handle only a certain type; we do have some workers which are designated as backlog workers, but those are for when we do a bulk index.  (I think they are designated as backlog workers differently as well; it's not by type.)

Obviously we could simply stop all the workers but that stops all indexing, even of the types that aren't being reindexed.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+
+.. code-block:: bash
+
+    GET kbase.genome_1/_mapping
+
+5b. Copy-paste the mapping from the current index into the body section of the PUT command below and make the necessary field change (preferably one change per complete reindexing operation).
+
+It is a good practice to make the mapping strict ("dynamic": "strict") for each type (data and access) in the index. Strict mappings prevent the mapping from being modified dynamically during ingest time.
+
+Update the settings section below the mapping. The number of shards and replicas must be decided on based on your capacity planning rules. It is costly to change the number of shards, so choose wisely! Make sure not to exceed 600 shards for any node in the system. Increase number of replicas to improve availability. In general,
```
kkellerlbl: We have not really done any sort of analysis of our sharding strategy.  I suspect we are just letting ES make all the decisions.  Do we need such an analysis?  (And is it relevant to updating an index?)

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+    {
+     "actions": [
+     {
+       "add": {
+         "index": "kbase.genome_1_r1",
+         "alias": "kbase.genome"
+         }
+       },
+       {
+         "remove": {
```
kkellerlbl, line 329: Is removing the old index even required if we're using aliases in this way?

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+
+    GET /_cat/aliases/kbase.genome
+
+    GET /_cat/indices/kbase.genome_*
+
+11. If the change involved in the reindexing operation also requires a corresponding search type spec change (located in resources/types/genome.yml for example), then this change must be applied to the spec as well.
+
+12. Change mapping version from "1" to "_r1" in the resources/types/genome.yml search type spec and add a comment (for future reference) that describes the change that took place in the r1 reindexing operation.
+
+13. Generate new events for the time range for which the change was made to the data source.
```
kkellerlbl: Docs for this?

kkellerlbl (changes requested):
A few questions in my review.

docsource/ops.rst
arfathpasha: I am not really seeing much benefit from having the '_r' versioning tag now. I thought initially it would help for tracking reindexed indexes through the process if more than one reindexing operation was being performed at once, but it also adds an extra step and code change. Given that reindexing will likely be done on one index at a time and atomically, I have removed this tag and the steps now move the index from version _1 directly to _2 instead of _1 -> _1_r1 -> _2

docsource/ops.rst
arfathpasha: Capturing slack conversation here. It is likely that indexing can be done within a few hours or at most overnight and this would be an acceptable amount of time. If we find through experimentation on CI that it will take much longer, a fallback option is to set the index that is being reindexed into read-only mode, then perform reindexing and update all failed events to unproc before restarting the trickle updates.

docsource/ops.rst
arfathpasha: It's not relevant to reindexing but it is important to have a separate capacity plan as it involves the determination of the number of ingest nodes, data nodes etc among other things. I think we should take this up as a separate task and work it out based on current needs and our projected growth. Since a lot of decision making in ES starts with "it depends" it might be better to start with defaults, see how the system responds and then use that baseline at least to determine current needs. I have documented limits here for #shards and shard size as a means to double check if we are within reasonable bounds.

docsource/ops.rst
arfathpasha: We don't have to and I put it here only as a best practice to start with. At some point we will have indexes for about a hundred different types and we would need some tooling in order to automatically diff all indexes of a type against all alias references for that type and then delete the unreferenced ones (probably oldest) when we run low on space. Until we have such tooling, it might help to follow some good house keeping practices.

docsource/ops.rst
arfathpasha: This is a todo item for now. I will fill it in once the CLI has been developed and merged.

docsource/ops.rst
MrCreosote: > It's not relevant to reindexing

Naively it seems like it is since you'll have to create a new index and you'll want to create it with the correct number of shards.

docsource/ops.rst
MrCreosote: This is just removing the index alias, not the index itself, right? So it's just flipping the target of the alias?

If so this is necessary because you can't write to an alias that points to more than one index IIRC. Also, presumably you don't want to search both the old and the new index.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+For each type of index, "genome" for example, we may have multiple indices in ElasticSearch with the naming convention "kbase.genome_1", "kbase.genome_2" etc. where the namespace "kbase" comes from the "elastic-namespace" key-value pair in search_tools.cfg and each numeric suffix comes from the search type version (See resources/typemappings/GenomeAndAssembly.yaml.example).
+
+In order to make it easy for the client to search across all indices of a certain type, aliases are applied with the following naming convention.
+
+.. code-block:: text
+
+ alias -> [indices]
+ genome -> [genome_1, genome_2 ... genome_n]
+
+When reindexing is necessary, the index is reindexed to a new index with a new search type version. The first reindexing operation performed on genome_1 will result in a new index genome_2 that replaces genome_1 in the genome alias.
```
MrCreosote, line 58: If I'm not misunderstanding, this seems to be conflating two different things - in the current implementation, the _* indicates the version of the search type (e.g. the index of the versions array in the spec yaml document, for example https://github.com/kbase/KBaseSearchEngine/blob/master/resources/types/Assembly.yaml#L4). This line seems to be proposing that _* also indicates the reindexing count.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+For the sake of simplicity and for the reasons described below, a single process has been defined for all reindexing cases -
+
+a) change field value
+b) add field type
+c) change field type
+d) remove field type
+
+The process involves reindexing an existing index into a new index for all of these cases. i.e. in-place reindexing is discouraged because the system may not have a recent snapshot/backup or any snapshot for recovery purposes should anything go wrong with the reindexing process. It is recommended that all of these steps are performed atomically for one index at a time (avoid reindexing multiple indexes in parallel) since it will be difficult to track through hundreds to indexes to find out which indexes are being reindexed.
+
+In addition, given that there can be as many as a thousand indices for KBase data, maintenance can become a challenge if the process is not simple. Some level of simplicity has been achieved here by defiining a single process that covers all the reindexing cases. If necessary, the process may be further simplified through some level of automation as it matures over time.
```
MrCreosote: defiining

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+
+    The commands below can be copy-pasted into Kibana and executed against the index. The corresponding curl commands can be obtained from Kibana by clicking on the little wrench icon that appears next to the pasted command.
+
+.. note::
+
+    If any of the steps below fail, don't proceed until the issue is resolved by referring to the ElasticSearch documentation.
+
+1. Note current time
+
+2. Stop any workers that are performing trickle updates on the index that needs to be reindexed. Note that this will stop the trickle updates to all indexes.
```
MrCreosote, line 164: I still think that this is going to bite us - every time we need to reindex Genomes search will stop updating for days. I know @scanon didn't want to set up a parallel indexing pipeline but IMO we need to rethink that.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+
+.. code-block:: bash
+
+    GET kbase.genome_1/_mapping
+
+5b. Copy-paste the mapping from the current index into the body section of the PUT command below and make the necessary field change (preferably one change per complete reindexing operation).
+
+It is a good practice to make the mapping strict ("dynamic": "strict") for each type (data and access) in the index. Strict mappings prevent the mapping from being modified dynamically during ingest time.
+
+Update the settings section below the mapping. The number of shards and replicas must be decided on based on your capacity planning rules. It is costly to change the number of shards, so choose wisely! In general, follow these rules and limits.
```
MrCreosote: I would just say changing the number of shards will require reindexing.

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+      "source": {
+        "index": "kbase.genome_1"
+      },
+      "dest": {
+        "index": "kbase.genome_2",
+        "version_type": "external"
+      },
+      "script": {
+        "lang": "painless",
+        "inline": "if (ctx._source.foo == 'bar') {ctx._version++; ctx._source.remove('foo')}"
```
MrCreosote, line 251: Will this operate on every object inserted into the index, even after reindexing is complete? If so, should it be shut off after reindexing?

docsource/ops.rst
```diff
@@ -1,6 +1,236 @@
...
+        "query": {
+          ...
+        }
+      },
+      "dest": {
+        "index": "kbase.genome_2"
+      }
+    }
+
+7. Run a checksum on the new index to make sure the numbers line up with the numbers of the current index.
```
MrCreosote, line 284: I'm not clear on how the checksums work but if we're making changes to the docs wouldn't we expect the checksums to be different?

MrCreosote, line 284: I'm still not seeing instructions for the case where you need to reindex from the source data (adding a field to the spec, for example).

MrCreosote (changes requested):

docsource/ops.rst
MrCreosote: E.g. 
```
gavin@login1:~$ curl ci-elasticsearch.kbase.us:9200/_cat/indices?v
health status index                                     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   kbaseci.genome_2                          IDnemxcQRjmFUy_8OqMN5w   5   0      52982            0     91.7mb         91.7mb
green  open   kbaseci.tree_1                            5k2IY-CcSoGVAJxcF5U4tw   5   0        290            8      515kb          515kb
green  open   kbaseci.assemblycontig_1_sub              biN0f0nFQk-GEIybrnbt5A   5   0   23378937      1610718      6.8gb          6.8gb
green  open   kbaseci.media_1                           RaLTgm8FQi-Ulyb5WitBvw   5   0       2604            6      1.1mb          1.1mb
green  open   kbaseci.singleendlibrary_1                -7bVLtEZTCGGBaec1S6LMA   5   0       2686            1      1.2mb          1.2mb
green  open   kbaseci.mediacompound_1_sub               whjdSCjCSuWpJ3YCEIjbSQ   5   0      28525           85      8.3mb          8.3mb
green  open   kbaseci.genome_1                          ZC2RXqqGTCaiBMBw5XUkzw   5   0     186410          145    107.3mb        107.3mb
green  open   kbaseci.pangenomeorthologfamily_1_sub     AWpAYj4oQHuNdgYe1XBsQQ   5   0     981447        62750    297.8mb        297.8mb
green  open   kbaseci.modelreaction_1_sub               wI9TmjaJR4i2p3PYqtkcYw   5   0    1511792        13725    363.1mb        363.1mb
```

arfathpasha: Here's my view. There are two ways to make a change to an index. 
1. Update the spec by creating a new set of indexing rules and add a mapping with a new version number (say 2). This would result in two indexes genome_1 (existing) and genome_2 (new). 
2. Reindex genome_1 into genome_2 and then update the spec by creating a new set of indexing rules (that match up with the reindexed changes, if any) and add a mapping with a new version number (genome_2). This would also result in two indexes genome_1 (existing) and genome_2 (new). 

So in this view, both methods help move indexes from one version to the next and are therefore synonymous in this respect. Maintaining a reindex count, as I previously thought was useful, I now feel serves no value (as long as these steps are performed atomically) and only adds more steps.

docsource/ops.rst
arfathpasha: Fixed.

docsource/ops.rst
arfathpasha: Done.

docsource/ops.rst
arfathpasha: Not sure I follow. The script is part of the POST _reindex call which is a one time call. This script will never be used (say for trickle updates) once the new index has been created and populated through the reindexing operation.

docsource/ops.rst
arfathpasha: This is just a sanity check. For example, if we make changes to the docs but don't add or delete docs, we should expect the doc count of the new index to be the same as the original index doc count, and size of the new index to be proportionally different from the size of the original index. If doc count has changed (say it drops) then it indicates that something didn't go right.

arfathpasha: Sorry, forgot to add this. I have a drafty version now from a couple of coarse runs. I plan to run these steps a few times next week to further refine, but this should be a good starting point.

docsource/ops.rst
scanon: While this may work, I think it would get confusing and it would mean we would have to leave old index specs in place just to keep the ordering.  I think I would prefer something like this...

1) Make a new index spec (say genome2.yaml).  This may include multiple versions if the ws type can't be handled by a single spec (e.g. the recent update to the genome object)

2) Add this new spec as an additional index in the mapping file. (as you described)

So it may look like...
    default:
           -
               type: Genome
               version: 2
           -
               type: Genome2
               version: 2
           -
               type: GenomeFeature
               version: 2 
etc...

So then we would see something like genome_1,genome_2 and genome2_1 and genome2_2.

I'll admit this may look a little confusing but I would be worried about mixing index spec versions in the same type file.

After things are checked out, then the alias is updated to use genome2_1 and genome2_2, and, eventually, the old versions are dropped from the mapping and the old indexes can be dropped.

arfathpasha:
Let me know if there is more to be done for this PR. I have already included the case that Shane requested - reindexing from source with a new type spec under the sub-heading "Reindexing from source"

scanon:
We are changing our strategy entirely so I'm going to close this.

PR #244, opened 2018-05-08
SCT-755 html documentation

arfathpasha:


MrCreosote:
This seems pretty straightforward. A couple questions:

* Are all the files in the PR required for the HTML docs or are the build artifacts remaining that could be deleted?
* Should the makefile (or whatever) be updated to build the HTML docs? If so, should that be part of the standard build or its own target?

arfathpasha:
There are probably a couple of artifact files like objects.inv. Since it is generated output, I would rather just treat is as a blackbox.

Makefile has the commands to build html and pdf outputs. The readme.md has the instructions for this.

PR #245, opened 2018-05-08
SCT-755  strict mapping

arfathpasha:
- added dynamic strict mapping
- fixed a bug with indexing returning successfully (bulk indexing returning status code 200) when some indexing operations fail


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -371,14 +374,46 @@ public void indexObjects(
...
+    private void verify(Response response) throws IOException {
+        // convert JSON string to Map
+        String responseJsonStr = EntityUtils.toString(response.getEntity());
+
+        ObjectMapper mapper = new ObjectMapper();
+        final Map<String, Object> responseMap = mapper.readValue(responseJsonStr,
+                new TypeReference<Map<String, Object>>() {});
+
+        // check for and collect response data from index operations that failed
+        List errorItems = new ArrayList();
```
MrCreosote: final

MrCreosote: Should there be a max size for this list? The exception text could be really big for a large load. Maybe log everything but only throw a portion of the list?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -371,14 +374,46 @@ public void indexObjects(
...
+        if(responseMap.get("errors").equals(Boolean.TRUE)) {
```
MrCreosote: space to distinguish key words from functions all through this part

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -371,14 +374,46 @@ public void indexObjects(
...
+            for(Map item: (List<Map>)responseMap.get("items")) {
+                for(Object itemData: item.values()) {
+                    if (!((Map) itemData).get("status")
+                            .toString().startsWith("20")) {  // status code 200, 201 etc
+                        errorItems.add(itemData);
```
MrCreosote: How big is itemData?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -371,14 +374,46 @@ public void indexObjects(
...
+                    }
+                }
+            }
+        }
+
+        // report response data from failed index operations
+        if(errorItems.size() > 0) {
+            throw new IOException("Indexing errors: " + errorItems.toString());
```
MrCreosote: Doesn't seem like there's any tests for this

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -629,8 +627,52 @@ public void pruneNarrative() throws Exception {
         assertThat("incorrect keyprops", od.getKeyProps(),
                 is(ImmutableMap.of("title", "a title")));
     }
-    
-    /* ****** Auth client tests - to be moved to their own suite *** 
+
+
+    @Test
```
MrCreosote: This seems like it'd be a better fit for a unit test rather than an integration test.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -23,6 +23,8 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.core.type.TypeReference;
```
MrCreosote, line 26: Might want to prevent dynamic index creation as well (not sure about this since it's cluster wide):
```
PUT _cluster/settings
{
 "persistent": {
   "action.auto_create_index" : false
 }
}
```

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -584,21 +585,18 @@ public void status() throws Exception {
     public void pruneNarrative() throws Exception {
         wsCli1.createWorkspace(new CreateWorkspaceParams()
                 .withWorkspace("narprune"));
-        
+
         final Map<String, Object> data = new HashMap<>();
         data.put("source", "a long string");
```
MrCreosote: Why is this test changing? Seems like the changes defeat the purpose of the test.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed.

arfathpasha: It won't be larger than the load itself so I am not sure if I should just put some arbitrary number. Logging would be great but at the moment there is no logging in ElasticIndexingStorage. I will create a separate PR for that.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Fixed.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Not very big. 3 or 4 key value pairs.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: SearchAPIIntegrationTest.strictMappingFail() covers this case.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
arfathpasha: I prefer the integration test because the verification code relies on the ES response. If I mock the ES error response and we change ES version in the future to a version that has a slightly different message, the unit test will silently pass each time and the code may become buggy.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: I'll have to look into this and make a separate PR for it. I have to think about how to do this as a one time operation for the entire index.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
arfathpasha: Because it was violating the strict mapping rule.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: The load size could be up to 100k docs - that's a really really large exception string. 

Logging should be easy - just use slf4j and logback classic (which may already be in the ant file) and do 

```
	private void logInfo(final String format, final Object... params) {
		LoggerFactory.getLogger(getClass()).info(format, params);
	}
	
	private void logErr(final String format, final Object... params) {
		LoggerFactory.getLogger(getClass()).error(format, params);
	}
```

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Ok, so the test really tests 2 separate things - can you note that in comments in the test?

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: Let me rephrase - this seems like it belongs in ElasticIndexingStorageTest (which aren't unit tests, strictly speaking) rather than the API integration test

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Rethinking this - there may be other applications using the same ES install that need dynamic index creation. Maybe just document it in setup that it's preferred but not required.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: Then shouldn't the test should be updated so it doesn't violate that rule but still tests what it's intended to test?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: That'll be automatically be intercepted by the built in logger in the SearchAPI and will just go to stdout for the workers, I believe.

arfathpasha: I have removed the exception and added a statement to log the errors.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Not sure I follow. I am only testing one assertion, which has now changed to verifying the log statement and I have added a comment for this.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
arfathpasha: Moved to ElasticIndexingStorageTest.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Added a note in ops.rst documentation in a separate PR

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
arfathpasha: That's exactly the change that I made.

scanon: I'm confused too. I was guessing the point of this test was to make sure that these other key values are dropped from the indexed object.  Why is it violating the strict mapping rule if they are getting dropped?

arfathpasha:
The "data" in the test is not raw data, but is supposed to simulate parsed data that has been extracted based on the indexing rules. On line 606, you can see that it is being used to create an instance of ParsedObject which is one of the arguments to the indexObjects() method. To make this clear, I have renamed "data" to "parsedData".

```
ImmutableMap.of(new GUID("WS:1/1/1"), new ParsedObject(
                         new ObjectMapper().writeValueAsString(data),
                         data.entrySet().stream().collect(Collectors.toMap(
                                 e -> e.getKey(), e -> Arrays.asList(e.getValue()))))),
                 false);
```

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
scanon: But this test is meant to test the pruning that happens in the search api.

Here
https://github.com/kbase/KBaseSearchEngine/blob/master/lib/src/kbasesearchengine/main/TemporaryNarrativePruner.java

So the idea, as I understand it, is the index record has these keys and we want to make sure the search api is dropping them.  So excluding them from the start doesn't test the pruner.

I'm still confused as to why the strict mapping is breaking things because those should be valid keys.

arfathpasha: I totally missed the documentation on top of the test about this test belonging to the narrative pruner. I have put back the keys that get pruned to show that the pruning functionality is working. To answer your question about why the strict mapping is breaking this test, it is because the keys in the parsed object need to be backed up with indexing rules as the indexing rules are used to create the mapping. No rule for a key, implies no mapping for that key and this causes the indexing to fail. So now, all I have done is add indexing rules for each of the keys.

scanon:
Looks good.

PR #255, opened 2018-06-13
SCT-1058 NarrativeInfo - Error handling for deleted workspaces in search results

ugswork:


eapearson:
Thanks, uma. 
I just wanted to take this opportunity to raise again a couple of issues.

First is how it is possible to get into this situation, in which search results are returned for deleted objects. It is expected that there is a lag between workspace changes and the index update, and in that window this condition may arise. 
But in the reported case, and others, the condition for the search result appears permanent.
I don't know if workspace events are considered to be lossy or not, but that is one possible explanation in this case; missed workspace events. Could also simple be a bug somewhere in workspace event emission, transport, handling, etc.

In the context of this change, perhaps the detection of this condition could lead to back pressure on the index to ensure that the offending index entires are corrected?

The other issue is how search represents this condition to callers of the searchapi. I think this set of changes will set the workspace info to null, which is consistent with how workspace calls handle ignoring errors. Another approach is to consider omitting all traces of invalid search results like this and return the results as if the inaccessible results were not there. It may be a simple statement, but it is a complex problem. What about summary results? They will include the invalid results because the summary (search_types) does not consult the container workspace. It also would require multiple ES queries to fill up the results buffer (whatever the page size is). But, it would present a consistent view of search. The alternative requires that all consumers of search account for results which may be associated with inaccessible objects, which could lead to cascading errors.

lib/src/kbasesearchengine/main/AccessGroupNarrativeInfoProvider.java
```diff
@@ -38,20 +38,19 @@ public AccessGroupNarrativeInfoProvider(final WorkspaceEventHandler wsHandler) {
      * @param wsid workspace id.
```
arfathpasha, line 40: Clarify documentation: 
For the given workspace ID, returns workspace info for the narrative, or null if no workspace info could coud be retrieved or found.

lib/src/kbasesearchengine/main/NarrativeInfoCache.java
```diff
@@ -79,12 +79,13 @@ public NarrativeInfo load(Long wsId) throws Exception {
     }
 
     @Override
-    public NarrativeInfo findNarrativeInfo(final Long wsId) throws IOException {
+    public NarrativeInfo findNarrativeInfo(final Long wsId) {
         try {
             return cache.get(wsId);
         } catch (ExecutionException e) {
-            throw (IOException) e.getCause(); // IOE is the only checked exception
+            System.out.println("ERROR: Failed retrieving narrative info: " + e.getMessage());
```
arfathpasha: To be consistent, use the logger
```
import org.slf4j.LoggerFactory;

    private void logInfo(final String format, final Object... params) {
        LoggerFactory.getLogger(getClass()).info(format, params);
    }

    private void logErr(final String format, final Object... params) {
        LoggerFactory.getLogger(getClass()).error(format, params);
    }
```

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -127,21 +126,34 @@ public GetObjectsOutput getObjects(final GetObjectsInput params, final String us
...
+            // defaults to the existing name so names from previous decorator layers aren't set to null
+            for (Map.Entry<Long, Tuple5<String, Long, Long, String, String>> entry : retVal.entrySet()) {
+                Tuple5<String, Long, Long, String, String> value = entry.getValue();
+                if (value != null) {
+                    value.withE5(displayNames.getOrDefault(value.getE4(), value.getE5()));
+                }
+            }
+        }
+        catch (IOException | Auth2Exception e) {
+            System.out.println("ERROR: Failed retrieving workspace owner realname(s):  " + e.getMessage());
```
arfathpasha: Use logger for consistency.

arfathpasha (commented):
some minor changes.

PR #227, opened 2018-03-15
SCT-635 Recursive indexing of objects public or private 

ugswork:

PR #256, opened 2018-06-18
SCT 1087 - AccessGroupsInfo/objectsInfo Error handling of deleted workspaces in search results

ugswork:
The accessGroupsInfo map added to the search results contain null value, for those workspaces (key: accessGroupId) which have been deleted.

The objectsInfo map added to the search results contain null value, for those objects (key: object reference) which have been deleted.


MrCreosote:
This appears to have changes that are not related to the purpose of PR - changes seem to be included from #227 for example. This needs to be separated into the changes that are required to prevent errors when workspaces are deleted, which I assume should be much less than 900 lines of code.

tiramisu24:
@MrCreosote @arfathpasha can this be closed and can one of you look at #261?

PR #257, opened 2018-06-18
SCT-1096: remove duplicate latest version genomes

arfathpasha:


scanon:
This looks good to me.

arfathpasha:
A caveat to note here is that I have used parentGUID for the parent object id (also used for routing). I did this to make it easy to type out queries like, 

```
GET assembly_1/data/WS:1_2_1?routing=WS:1_2_1
```

One problem with this is that all objects from a particular workspace will end up in the same shard and this can potentially cause shard size skewing if there are some workspaces that are extraordinarily large. If I revert to default ids for routing, it will resolve the shard skewing problem but I don't think it is possible to run this type of a query since the routing id will be some random string.

MrCreosote:
Remind me why the ID scheme above will result in all the objects for a particular workspace winding up on the same shard?

arfathpasha:
Here's the link that explains the routing. In short, "This parent ID serves two purposes: it creates the link between the parent and the child, and it ensures that the child document is stored on the same shard as the parent. The routing to a shard is done with a formula ```shard = hash(routing) % number_of_primary_shards```. If a parent ID is specified, it is used as the routing value."

https://www.elastic.co/guide/en/elasticsearch/guide/current/indexing-parent-child.html

MrCreosote:
So that seems to mean then that all objects in a workspace will *not* go to the same shard, because the parent ID will be hashed, right? And, for example, WS:3_4_1 will get a different hash, and therefore potentially a different shard, than WS:3_5_1?

It does mean that all the *subobjects* of a particular object version will go to the same shard. Naively that seems ok to me.

Am I missing something?

MrCreosote:
If we get rid of the access document (please please please) this'll be moot...

arfathpasha:
Right. That's what I meant - all subobjects will end up in the same shard. I remember you mentioned that some objects can have thousands of subobjects and I thought this would become a problem. Yes, I can work on dropping the parent-child and problem solved.

MrCreosote:
> Right. That's what I meant - all subobjects will end up in the same shard. I remember you mentioned that some objects can have thousands of subobjects and I thought this would become a problem.

They do, but there are hundreds of thousands to millions of objects like that, so as long as those objects are spread around the shards reasonably randomly, I suspect things should be ok.

Plants and mammals might be a bit more worrisome as there are fewer genomes and they have more subobjects, but even then I doubt it'd be bad enough to cause severely unbalanced shards.

> Yes, I can work on dropping the parent-child and problem solved.

I don't think we have agreement that we can do that yet, unfortunately.

MrCreosote:
Looking at the fix implementation, IIUC it's basically searching every extant index looking for objects with the correct parent GUID, and then loops over every discovered indexed to set the properties correctly.

Would a simpler implementation that just uses wild cards be possible? E.g. instead of using the index name that's returned by checkIndex() (or not using an index name at all), use the prior implementation but just have loadVersion run on, say, `kbaseci.genome_*`? Same thing for updating the versions. If that's possible, the code could be a lot simpler as the new implementation wouldn't need the loops/maps/etc. Also it'd only search indexes that can actually contain the necessary info, rather than all indexes.

arfathpasha:
Using wildcards was actually my first version of this PR that Shane thought was good. While it worked well for parent objects, I found it caused nasty conflict errors for sub-objects after running a test on CI. The scenario is as follows: Assume there are three genomeFeature sub-objects that need to be inserted. When subobject1 is inserted, it gets ES _version of 1. When subobject2 is inserted, subobject1 ends up with ES _version 2 and subobject2 gets ES _version of 1, and so on. So each ES _version of each of the subobjects is different and when I run a blanket update on all of them using the wildcard approach, I get _version conflict errors on every run. In the interest of time, I didn't get into the nitty-gritty of trying to find out exactly why the version conflict errors were happening, but I am guessing it is because of the lack of the ability to refresh (when using the wildcard approach) as multiple subobjects are being updated. But as you can see with the current implementation I was able to overcome these conflict errors by retrieving each of these objects from their respective indexes, getting their current ES _versions and then running the update using the current _version, and more importantly also applying a refresh for each update. If the refresh argument is taken out, the conflict errors happen occasionally, hence my guess that it could be an internal issue related to the ES records not being refreshed before the next update. Now, I know refresh is meant to make the record become available for search and by default it happens periodically, but as I explained here, it also seems to have an effect on updates.

arfathpasha:
I think I figured out the actual problem. The sequence is - 

1. _index, 
2. get ids of records whose last version field needs to be updated (done with a _search!)
3. _update field in records

The search in step 2 requires the refresh on every bulk indexing and update operation, else we get the version conflict exceptions. So, a simple refresh solves the problem and therefore I am going to close this PR in favor of the lighter change in the new [#PR/259](https://github.com/kbase/KBaseSearchEngine/pull/259)

PR #258, opened 2018-06-20
SCT-1107: Dev cli overwrite

arfathpasha:


lib/src/kbasesearchengine/events/StatusEvent.java
```diff
@@ -378,13 +400,27 @@ public Builder withNullableNewName(final String newName) {
             }
             return this;
         }
+
+        /** Specify whether the corresponding record for this event should be overwritten
+         * if the record already exists in the index."
+         * @param overwriteExistingData true if the existing record should be overwritten, else false.
+         * @return this builder.
+         */
+        public Builder withOverwriteExistingData(final Boolean overwriteExistingData) {
```
MrCreosote: I would call the method with withNullableOverwrite... to maintain consistency with the rest of the API and let the user know the boolean may be null.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -127,6 +127,7 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
                 .withNullableAccessGroupID(6)
                 .withNullableisPublic(true)
                 .withNullableNewName("foo")
+                .withOverwriteExistingData(true)
```
MrCreosote: Shouldn't these tests check that false and null are stored correctly?

lib/src/kbasesearchengine/events/StatusEvent.java
```diff
@@ -33,6 +33,7 @@
     private final Optional<Integer> version;
```
MrCreosote, line 33: Since the CLI and event generator aren't covered by tests I assume they were tested manually?

lib/src/kbasesearchengine/events/StatusEvent.java
```diff
@@ -125,6 +128,14 @@ public Instant getTimestamp() {
         return isPublic;
     }
 
+    /** Get whether the corresponding record for this event should be overwritten
+     * if the record already exists in the index."
+     * @return whether to overwrite existing data.
+     */
+    public Optional<Boolean> isOverwriteExistingData() {
```
MrCreosote: Actually, is there any need for an optional here? can the builder assume that null = don't overwrite?

test/src/kbasesearchengine/test/main/IndexerWorkerTest.java
```diff
@@ -1138,6 +1139,188 @@ public void skipEvent() throws Exception {
...
+                new StatusEventID("pid"));
+
+        when(ws.updateObjectEvent(event.getEvent())).thenReturn(event.getEvent());
+
+        // simulate to show that index record for corresponding event already exists
+        when(idxStore.checkParentGuidsExist(set(guid))).thenReturn(ImmutableMap.of(guid, true));
+
+        final StatusEventProcessingState res = worker.processEvent(event);
+        assertThat("incorrect state", res, is(StatusEventProcessingState.INDX));
+
```
MrCreosote, line 1319: Should be a check here that indexObjects() was called on the indexing storage system

test/src/kbasesearchengine/test/main/IndexerWorkerTest.java
```diff
@@ -1138,6 +1139,188 @@ public void skipEvent() throws Exception {
...
```
MrCreosote, line 1225: Should check that there was no interactions with the indexing storage system other than checking the parent guid

MrCreosote (changes requested):
Looks good, a few small changes.

lib/src/kbasesearchengine/events/StatusEvent.java
arfathpasha: Yes. I ran both against CI and they look good.

lib/src/kbasesearchengine/events/StatusEvent.java
arfathpasha: Yes, but I would rather keep it consistent throughout. i.e. the field type, input type and output type should be same for these pojos. If we decide to go with the convention of just true/false with false as default, then there is no need to use the Optional wrapper in any of the pojos.

test/src/kbasesearchengine/test/main/IndexerWorkerTest.java
arfathpasha: Added.

test/src/kbasesearchengine/test/main/IndexerWorkerTest.java
arfathpasha: added.

lib/src/kbasesearchengine/events/StatusEvent.java
MrCreosote: I think there's a difference here in that for isPublic(), it's possible that the event may not have that information available depending on the event type, so absent() indicates the info is not available and the service should not make any assumptions re whether the data is public or not.

For isOverwrite(), if the info is not available the choice is obviously going to be to not destroy data, so absent() and false have identical semantics, unlike isPublic(), which is why IMO there's no need for 3 states.

If your heart is set on having 3 states, then I think the minimum requirement here is to document the semantics of the absent() state.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -127,6 +127,7 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
                 .withNullableAccessGroupID(6)
```
MrCreosote, line 127: Copying comment from previous review:
Shouldn't these tests check that false and null are stored correctly?

MrCreosote (changes requested):

lib/src/kbasesearchengine/events/StatusEvent.java
arfathpasha: I clarified with additional documentation. "An absent field indicates that the record must not be overwritten."

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Added tests.

lib/src/kbasesearchengine/events/StatusEvent.java
MrCreosote: Which, to reiterate, means that the semantics of false and absent are identical and therefore absent is redundant, but *shrug*

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,72 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+                "WS", Instant.ofEpochMilli(10000), StatusEventType.COPY_ACCESS_GROUP)
+                        .withNullableisPublic(false)
+                        .withNullableOverwriteExistingData(false)
+                        .build(),
+                StatusEventProcessingState.UNPROC,
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect access group id", sse.getEvent().isPublic(), is(Optional.of(false)));
+        assertThat("incorrect access group id", sse.getEvent().isOverwriteExistingData(), is(Optional.of(false)));
```
MrCreosote: Is there a reason this doesn't test get() like the null fields test?

Also I realize that the tests don't actually test that the true fields work either. They store it but never check it as far as I can tell. Should probably add that.

After that i think this is done.

MrCreosote (changes requested):

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: I added a couple more test methods. So now storeAndGetNoTypeEmptyCodesAllFields() tests non -default values. storeAndGetAllBooleanFields() tests boolean values for both T/F. storeAndGetNullableDefaults() tests for defaults (which is Optional.absent())

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
         assertThat("incorrect error msg", got.getErrorMessage(), is(Optional.absent()));
         assertThat("incorrect error trace", got.getErrorStackTrace(), is(Optional.absent()));
     }
+
+    @Test
+    public void storeAndGetAllBooleanFields() throws Exception {
```
MrCreosote, line 184: These tests should pull the event back from the db and check the results

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+        final StoredStatusEvent sse1 = storage.store(StatusEvent.getBuilder(
+                "WS", Instant.ofEpochMilli(10000), StatusEventType.COPY_ACCESS_GROUP)
+                        .withNullableisPublic(false)
+                        .withNullableOverwriteExistingData(false)
+                        .build(),
+                StatusEventProcessingState.UNPROC,
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect access group id", sse1.getEvent().isPublic(), is(Optional.of(false)));
```
MrCreosote: Looks like the assert messages need to be updated from the C&P

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+                        .build(),
+                StatusEventProcessingState.UNPROC,
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect state", sse.getState(), is(StatusEventProcessingState.UNPROC));
+        assertThat("incorrect updater", sse.getUpdater(), is(Optional.absent()));
+        assertThat("incorrect update time", sse.getUpdateTime(), is(Optional.absent()));
+        assertThat("incorrect access group id", sse.getEvent().getAccessGroupId(), is(Optional.absent()));
+        assertThat("incorrect access group id", sse.getEvent().isPublic(), is(Optional.absent()));
```
MrCreosote: Looks like the assert messages need to be updated from the C&P

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+        // tests all nullable defaults
+        when(clock.instant()).thenReturn(Instant.ofEpochMilli(30000L));
+        final StoredStatusEvent  sse = storage.store(StatusEvent.getBuilder(
+                "WS", Instant.ofEpochMilli(10000), StatusEventType.COPY_ACCESS_GROUP)
+                        .build(),
+                StatusEventProcessingState.UNPROC,
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect access group id", sse.getEvent().getAccessGroupId(), is(Optional.absent()));
```
MrCreosote: Looks like the assert messages need to be updated from the C&P

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect access group id", sse2.getEvent().isPublic(), is(Optional.of(true)));
+        assertThat("incorrect access group id", sse2.getEvent().isOverwriteExistingData(), is(Optional.of(true)));
+    }
+
+    @Test
+    public void storeAndGetNullableDefaults() throws Exception {
+        // tests all nullable defaults
```
MrCreosote: Test should pull the event back from the db and check results

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
```diff
@@ -176,6 +179,104 @@ public void storeAndGetNoTypeEmptyCodesAllFields() throws Exception {
...
+                StatusEventProcessingState.UNPROC,
+                set(),
+                "Baldrick");
+
+        assertThat("incorrect access group id", sse2.getEvent().isPublic(), is(Optional.of(true)));
+        assertThat("incorrect access group id", sse2.getEvent().isOverwriteExistingData(), is(Optional.of(true)));
+    }
+
+    @Test
+    public void storeAndGetNullableDefaults() throws Exception {
```
MrCreosote: Actually, this test looks redundant to the test below

MrCreosote (changes requested):

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Done.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Fixed.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Fixed.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Fixed.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Fixed.

test/src/kbasesearchengine/test/events/storage/MongoDBStatusEventStorageTest.java
arfathpasha: Removed.

PR #259, opened 2018-06-29
SCT-1096: remove duplicate latest version genomes

arfathpasha:
2nd attempt! Verified on CI with ref genomes.


MrCreosote:
300 LOC -> 50 LOC for the fix, nice.

MrCreosote:
I still don't quite understand why going across more than one index makes the refresh necessary. Each object should be in one and only one index, so why would adding another index cause conflicts?

It seems possible the conflicts were happening because the objects would be inserted and then the islast field updated before a refresh occurred, but I don't understand why that wouldn't happen without the wildcarded indexes.

MrCreosote:
Do you think wait_for might be a better option? 
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -1733,7 +1738,7 @@ public Response makeRequestBulk(
             final File jsonData) 
             throws IOException, IndexingConflictException {
         try (InputStream is = new FileInputStream(jsonData)) {
-            return makeRequest(reqType, "/" + indexName + "/_bulk", Collections.emptyMap(),
+            return makeRequest(reqType, "/" + indexName + "/_bulk?refresh", Collections.emptyMap(),
```
MrCreosote: Should document that this method now causes an immediate refresh. AFAICT it's only used in indexObjects() so the change doesn't affect any other operations.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -242,6 +242,50 @@ private static ObjectData getIndexedObject(GUID guid) throws Exception {
...
+     * two indexes that exist for the type genome, where genome_1 contains records for some genome
+     * type version set x and genome_2 contains records for some genome type version set y, and the
+     * latest genome object version resides in, say genome_1, then only this record's islast field
+     * must be set among all records contained in genome_1 and genome_2. i.e. all records in genome_2
+     * must have their islast field set to false in this case.
+     *
+     */
+    @Test
+    public void testLastVersionForParentObjects() throws Exception {
+        indexObject("Genome", 0, "genome01", new GUID("WS:1/1/1"), "MyGenome.1");
```
MrCreosote: I assume that without the changes this test will fail because 1/1/2 will also be returned?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -242,6 +242,50 @@ private static ObjectData getIndexedObject(GUID guid) throws Exception {
...
+    public void testLastVersionForSubObjects() throws Exception {
+        indexObject("GenomeFeature", 0, "genome01", new GUID("WS:1/1/1"), "MyGenome.1");
+        indexObject("GenomeFeature", 0, "genome01", new GUID("WS:1/1/2"), "MyGenome.1");
+        indexObject("GenomeFeature", 1, "genome01", new GUID("WS:1/1/3"), "MyGenome.1");
+        indexObject("GenomeFeature", 1, "genome01", new GUID("WS:1/1/4"), "MyGenome.1");
+
+        MatchFilter filter = MatchFilter.getBuilder().build();
+        AccessFilter accessFilter = AccessFilter.create().withAllHistory(false).withAdmin(true);
+
+        FoundHits hits = indexStorage.searchObjects(Arrays.asList("GenomeFeature"), filter, null, accessFilter, null, null);
```
MrCreosote: line wrap

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -242,6 +242,50 @@ private static ObjectData getIndexedObject(GUID guid) throws Exception {
...
+        Assert.assertEquals("expected 3 guid", 3, hits.guids.size());
```
MrCreosote: Sorry, I'm confused here. Why are 3 objects expected in the return?

MrCreosote (changes requested):
Wound up making single comments, sorry.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -637,6 +637,11 @@ private int updateLastVersionsInData(String indexName, GUID parentGUID,
         if (indexName == null) {
```
MrCreosote: Actually should note that this method now forces a refresh as well. This is used (AFAICT) in indexObjects and undeleteAllVersions.

MrCreosote: Also, indexObjects does a refresh as its last operation, which is now unnecessary due to the updateVersion refresh.

MrCreosote: Another question - indexObjects uses checkIndex() to get the index name, which returns an index name with a version. Then it calls loadLastVersion() with that index name. That seems to imply that the last version loaded could be incorrect if the higher version number was in an index different from that to which the data was eventually indexed.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Done.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: That is correct.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Done.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
arfathpasha: Genome01 has 3 features.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Added comment with reason for the refresh.

arfathpasha: The next few lines after the referenced line above replace the version with a wild card in order to update all indexes with the specified index name.

arfathpasha:
Regarding the need for reindexing, I think my last comment on the closed PR link below explains why. The reason is not related to going across more than one index, but the process that is followed to make each update - 

https://github.com/kbase/KBaseSearchEngine/pull/257

I also looked at waitFor. This arg would make the indexer wait for the periodic refresh. Overall, both seem the same but forcing the refresh would save a few milliseconds for each index operation.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: Removed.

MrCreosote: That's not what I mean. The steps are:

1) indexObjects calls checkIndex
2) checkIndex returns an index name like MyType_2, where 2 is the version of the search type MyType
3) indexObjects calls loadLastVersion()
4) loadLastVersion() looks up the maximum version of the object in index MyType_2 and returns it

that's where we run into trouble, because there's no guarantee that the last version of the object is in the MyType_2 index. It might be in MyType_1, MyType_3, or whatever. So loadLastVersion() might return the wrong maximum version, which would cause islast to be set incorrectly when the versions are updated after the objects are saved.

MrCreosote:
> I also looked at waitFor. This arg would make the indexer wait for the periodic refresh. Overall, both seem the same but forcing the refresh would save a few milliseconds for each index operation.

Given that indexing, say, a genome or assembly takes several seconds, IIRC, it seems like the few milliseconds may be worth it for reducing the number of refreshes, which can actually speed up indexing performance:
https://sematext.com/blog/elasticsearch-refresh-interval-vs-indexing-performance/

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
MrCreosote: right, duh

MrCreosote:
I'm sorry, but the refresh stuff is still confusing to me.

> Regarding the need for reindexing, I think my last comment on the closed PR link below explains why. The reason is not related to going across more than one index, but the process that is followed to make each update -

> <copied from #257>:
>> I think I figured out the actual problem. The sequence is -
>>
>>   1. _index,
>>   2.  get ids of records whose last version field needs to be updated (done with a _search!)
>>   3. _update field in records
>>
>> The search in step 2 requires the refresh on every bulk indexing and update operation, else we get the version conflict exceptions. So, a simple refresh solves the problem

I don't think that's the sequence is it? From reading the code, it's

1. look up the last version
2. index the objects,
3. _update_by_query, which searches and updates the version in one shot.

I'm not seeing where a _search is used to find the objects, and then a separate _update is used to update those objects.

Assuming I'm correct, the only place where there could possibly be conflicts, AFAICT, is on step 3, where the objects that were indexed could be modified before refresh by _update_by_query. But that's the way the code has always worked - so why would these changes introduce version conflicts? The actual steps where version conflicts are possible don't change, as far as I can tell. That's why I'm confused.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
arfathpasha: I think you are right. I have changed loadLastVersion() to return the global last version and not the local last version.

arfathpasha:
In my sequence, both steps 2&3 take place in that single _update_by_query call. i.e. first there is a query that requires a search (step 2) and then comes the update (step 3)

arfathpasha:
I added wait_for to the refresh.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: Should add a test that exercises this scenario - I think just indexing a high object version in with a search spec with version X and then a low object version with a search spec with version Y will do it. Without the fix you'd expect the low version to be set with islast=true.

MrCreosote:
> In my sequence, both steps 2&3 take place in that  single _update_by_query call. i.e. first there is a query that requires a  search (step 2) and then comes the update (step 3)

Ok, so then I'm still confused why the changes in this PR would cause conflict errors.

> I added wait_for to the refresh.

Should this one get changed too? https://github.com/kbase/KBaseSearchEngine/pull/259/files#diff-91f54c4aa40cccd5db0523228216fcfdR1776

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
MrCreosote: You might also want to make the index name munging code a function, as it looks like it's duplicated in two places.

arfathpasha: Done.

MrCreosote:
What's still open on this one AFAICT are:

* Adding a test that exercises the version updating code bug you fixed.  I think just indexing a high object version in with a search spec with version X and then a low object version with a search spec with version Y will do it. Without the fix you'd expect the low version to be set with islast=true.

* I still don't understand why running getGeneralIndexName() (which is the only real change here) on the index name when running the version update necessitates a refresh. It's the exact same operation as before except it runs on, potentially, more than one index.

arfathpasha:
getGeneralIndexName() is a refactoring that you requested for similar code that was being used in two places and not a fix. 

I have reversed the indexing order as you suggested from high obj version to low obj version from two different spec versions and the tests continue to pass. 

The conflict errors are a subtle problem and the best way to make sense of them I think is to run the last version test especially for sub-objects without the refresh=wait_for argument and see the error output.

MrCreosote:
Tests look good.

> getGeneralIndexName() is a refactoring that you requested for similar code that was being used in two places and not a fix.

What I was trying to get at is that the fix is replacing the old index name retrieval code with `getGeneralIndexName()`, and that function contains the entirety of the fix. That's the only change in the diff other than `wait_for`.

Just to make sure I understand this, given 3 scenarios:

a) The codebase without this PR
b) The codebase with this PR but without the wait_for changes
c) The codebase with this PR in its entirety

Only scenario b) will result in conflict errors?

arfathpasha:
That is correct. If you remove refresh=wait_for you will get conflict errors (case b). The other two cases don't result in conflict errors, but case a has the duplication problem that case c solves.

MrCreosote:
that's bizarre and I don't understand it, but...

PR #236, opened 2018-04-18
SCT 720 - changes and tests for unauthenticated search

ugswork:

PR #260, opened 2018-07-03
New720

tiramisu24:
-cleanup
- fixed test in SearchAPIIntegrationTest


tiramisu24:
@MrCreosote can you look at this? Its based on pull request 236. I fixed the test in SearchAPI.

MrCreosote:
Replaces https://github.com/kbase/KBaseSearchEngine/pull/236

MrCreosote:
Per our conversation:

1) Update spec descriptions to clarify unauthed search
2) Test the untested paths though the if boolean (see codecov report)

Then I review.

tiramisu24:
@MrCreosote can you review

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
```diff
@@ -212,11 +212,14 @@ public void timeStat(GUID guid, long loadMs, long parseMs, long indexMs) {
      * @param   params   instance of type {@link kbasesearchengine.SearchTypesInput SearchTypesInput}
      * @return   instance of type {@link kbasesearchengine.SearchTypesOutput SearchTypesOutput}
      */
-    @JsonServerMethod(rpc = "KBaseSearchEngine.search_types", async=true)
+    @JsonServerMethod(rpc = "KBaseSearchEngine.search_types", authOptional=true, async=true)
     public SearchTypesOutput searchTypes(SearchTypesInput params, AuthToken authPart, RpcContext jsonRpcContext) throws Exception {
         SearchTypesOutput returnVal = null;
         //BEGIN search_types
-        returnVal = search.searchTypes(params, authPart.getUserName());
+        if (authPart != null)
```
MrCreosote: Could make this a method rather than duplicating 3 times.

something like 
```
private String getUserName(final AuthToken token) {
    return token == null ? null : token.getUserName();
}
```

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -282,10 +283,13 @@ public SearchObjectsOutput searchObjects(SearchObjectsInput params, String user)
     @Override
     public GetObjectsOutput getObjects(final GetObjectsInput params, final String user)
             throws Exception {
-
+        Set<Integer> accessGroupIDs;
```
MrCreosote: final

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -10,6 +10,7 @@
 import static org.mockito.Mockito.when;
 
 import java.io.File;
+import java.io.FileNotFoundException;
```
MrCreosote: Unused?

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -178,7 +179,7 @@ public void cleanup() throws Exception {
     @AfterClass
     public static void teardown() throws Exception {
         if (es != null) {
-            es.destroy(TestCommon.getDeleteTempFiles());
+                es.destroy(TestCommon.getDeleteTempFiles());
```
MrCreosote: Indenting looks wrong here

.gitignore
```diff
@@ -1,14 +1,21 @@
 test_local
```
MrCreosote: Needs release notes

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +671,32 @@ private void compare(final SortingRule got, final SortingRule expected) {
                 is(expected.getIsObjectProperty() == null ? 1L : expected.getIsObjectProperty()));
     }
 
+    @Test
```
MrCreosote: I'm confused about what this test is trying to do. Indexing storage is mocked, but no stubs are ever created, and so always returns the default value, meaning the test will pass even if the code is broken. The mocks should actually return something to show the calls are working, and various cases exercising all the paths through the SearchMethods code should be tested (usually 1 case per test unless there's a ton of setup code that's hard to stick in a function.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,130 @@ public void narrativeDecoration() throws Exception {
         NarrativeInfoDecoratorTest.compare(res.getAccessGroupNarrativeInfo(), expected);
     }
 
+    @Test
+    public void optionalAuthForSearch() throws Exception {
```
MrCreosote: This test is waaaaaaay too long, and, I presume, tests many assertions. It needs to be broken up into smaller tests, potentially with set up code reused in a method.

Another possibility is that if the unit tests for SearchMethods are complete, this test could probably be reduced to just testing regular and unauthed search integration (assuming it's now testing more than that).

MrCreosote (changes requested):

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
tiramisu24: it is just testing regular and unauthed search though

MrCreosote: There's two tests right there, in that case - and why is it 120 lines of code do just do two tests?

tiramisu24: everything before the try catch is setting up the workspace and indexing objects

tiramisu24: try catch is doing the calls, and everything after is checking it got the right results. Uma also indents alot so it looks long

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: fixed it

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: What I'm trying to get at is that the test should be as easy as possible to read and understand - we just had this problem yesterday with a test neither of us could understand, so you should get where I'm coming from :p

To make tests easy to understand they should 
1) be as short as possible
2) have a name describing what being tested
2a) with extra documentation if necessary
3) Only test one thing or concept or whatever. There are some people that say each test should have a single assert statement, but I don't go that far. IMO each test should test one action - like one call to a method. That might take multiple asserts to test thoroughly, but the point is that one method call tested per test is a good general rule.

If there's a lot of set up code that's shared between two method call tests, put it it its own method with a descriptive name.

4) Favor unit tests over integration tests. They're faster, easier to understand, and easier to debug. Integration tests should just make sure the basic paths through the integrated objects work and the object's message passing works. Unit tests should exercise the bulk of the code.

scanon: Let's not worry about trying to tighten this up further.  I really want to get this merged and running in CI so Erik can start testing it.

MrCreosote: It should take a few minutes to fix. We're already having this problem where apparently tests aren't preventing bugs from being introduced, or the tests we're adding aren't actually testing the code properly. The fix for that is tests that are clearer so it's easier to reason about them and figure out when they're not working.

tiramisu24: @MrCreosote I am not sure what I can take out here though that would make this easier to read. It follows the same pattern as all the other tests.

MrCreosote: You'd have separate tests for the 3 methods and 2 access modes, so 6 tests instead of one, each with descriptive names, and you'd put the setup code in a common method.

scanon:
Can we merge?

MrCreosote:
@scanon sorry, got sidetracked. Reviewing now.

.gitignore
MrCreosote: Needs release notes

.gitignore
```diff
@@ -1,14 +1,21 @@
...
+data
```
MrCreosote: Why is data excluded? Thats a standard SDK dir.

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
```diff
@@ -212,15 +212,18 @@ public void timeStat(GUID guid, long loadMs, long parseMs, long indexMs) {
...
     public SearchTypesOutput searchTypes(SearchTypesInput params, AuthToken authPart, RpcContext jsonRpcContext) throws Exception {
         SearchTypesOutput returnVal = null;
         //BEGIN search_types
-        returnVal = search.searchTypes(params, authPart.getUserName());
+        returnVal = search.searchTypes(params, getUserName(authPart));
         //END search_types
         return returnVal;
     }
 
+    private String getUserName(final AuthToken authPart){
```
MrCreosote: This will be lost the next time the server is recompiled. Remember all code has to go between //BEGIN and //END markers.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+            System.out.println("Exception server side trace:\n" + e.getData());
+            throw e;
+        }
+
+        assertThat("incorrect search objects count", noAuthSearchObjsOut.getObjects().size(), is(1));
+        TestCommon.compare(noAuthSearchObjsOut.getObjects().get(0), expected2);
+
+        assertThat("incorrect search objects count", authSearchObjsOut.getObjects().size(), is(2));
+
+        Set<String> res = new HashSet<>();
```
MrCreosote: Something like this is less convoluted:
```
final Set<String> got = authSearchObjectsOut.getObjects().stream()
    .map(o -> o.getGuid()).collect(Collectors.toSet);
assertThat("incorrect results", got, is(set("WS:1/1/1", "WS:2/1/1")));
```

Then you don't need the size check either.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+                Instant.ofEpochMilli(10000),
+                null,
+                new GUID("WS:2/1/1"),
+                ImmutableMap.of(new GUID("WS:2/1/1"), new ParsedObject(
+                        "{\"whee\": \"imaprettypony1\"}",
+                        ImmutableMap.of("whee", Arrays.asList("imaprettypony1")))),
+                true);  // isPublic set to true
+    }
+
+    private ObjectData setUpOptionalAuthForSearchGetRes(String guid, String objName) throws Exception{
```
MrCreosote: final args

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+                .withHighlight(new HashMap<>());
+    }
+    @Test
+    public void optionalAuthSearchObjects() throws Exception{
+        setUpOptionalAuthForSearch();
+
+        final ObjectData expected1 = setUpOptionalAuthForSearchGetRes("WS:1/1/1","objname1");
+        final ObjectData expected2 = setUpOptionalAuthForSearchGetRes("WS:2/1/1","objname2");
+
+        SearchObjectsOutput authSearchObjsOut;
```
MrCreosote: final

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+        res.add(expected2.getGuid());
+        assertThat("incorrect results", res.contains(authSearchObjsOut.getObjects().get(0).getGuid()), is(true));
+        assertThat("incorrect results", res.contains(authSearchObjsOut.getObjects().get(1).getGuid()), is(true));
+
+    }
+    @Test
+    public void optionalAuthForSearchTypes() throws Exception {
+        setUpOptionalAuthForSearch();
+
+        SearchTypesOutput authSearchTypesOut;
```
MrCreosote: final

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+        assertThat("incorrect search types map size", typeToCount.size(), is(1));
+        assertThat("Incorrect search types type name", typeToCount.keySet(), is(set("Deco")));
+        assertThat("Incorrect search types type count", typeToCount.get("Deco"), is(Long.valueOf(2)));
+    }
+
+    @Test
+    public void optionalAuthForGetObjects() throws Exception {
+        setUpOptionalAuthForSearch();
+
+        GetObjectsOutput authGetObjsOut;
```
MrCreosote: final

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+
+            authGetObjsOut = searchCli.getObjects(new GetObjectsInput()
+                    .withGuids(Arrays.asList("WS:1/1/1", "WS:2/1/1")));
+
+        } catch (ServerException e) {
+            System.out.println("Exception server side trace:\n" + e.getData());
+            throw e;
+        }
+
+        assertThat("incorrect get objects count", noAuthGetObjsOut.getObjects().size(), is(0));
```
MrCreosote: Since one of the objects is public shouldn't there be one object returned? Seems like a bug...

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
```diff
@@ -551,6 +566,232 @@ public void narrativeDecoration() throws Exception {
...
+        assertThat("incorrect get objects count", authGetObjsOut.getObjects().size(), is(2));
+
+    }
+    @Test
```
MrCreosote: I think this is the old test, right, and so should be deleted?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,58 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+                .withAccessFilter(new AccessFilter().withWithPrivate(1L))
+                .withMatchFilter(new MatchFilter());
+
+        //authorized and showing private data
+        final SearchTypesOutput res = sm.searchTypes(input1,
+                "auser");
+
+        assertThat("incorrect res", res.getTypeToCount(), is(expectedRes));
+
+        //unauthorized user
```
MrCreosote: These next two tests (which should be in separate tests) will always pass, AFAICT, because they depend on the default returns from Mockito. The only way they could fail is if they happened to trigger the stub set up above.

MrCreosote (changes requested):

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: so the last one is basically checking that the stub is not being triggered. If I add another stub for the no user case, that should be it?

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
tiramisu24: ops

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
tiramisu24: what then how do I do helper functions

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
MrCreosote: I would return *something* for all cases to be sure the test is actually working. If the third case has false public *and* false private... well that's kind of dumb. Maybe that should throw an exception, since it makes no sense as input to the function?

Also shouldn't there be 4 cases, since there's 2 booleans?

Also now I think of it shouldn't there be tests for get objects?

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
MrCreosote: Put them in the class header

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
tiramisu24: so getObjects lets you get objects you have access to (it checks if you have the right access group id and doesnt care about public/private). Is that a feature or bug?

tiramisu24: Also it unclear to me whats the usage of GetObjects if there is SearchObjects

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: The cases I see are: 
authorized user and showing private data
authorized user and not showing private data
unauthorized user (private data is hard coded as false here)

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: For get objects there's no way to specify private/public access, so you're right. The purpose of get objects is to allow getting objects directly by the guid. I assume it's a special case of search objects. It's weird in that it allows for a match filter *and* a guid list.

MrCreosote: Actually looking at the code, the get_objects match filter is ignored. Nice. That should be taken out of the spec.

MrCreosote: Wait a sec - if an unauthed user calls get_objects with a public guid they should definitely get it, so this does seem like a bug.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
MrCreosote: There's 5 cases I think:
Unauthorized (private = false, public = true) <- hardcoded booleans
Authorized with:
private = false, public = true
private = true, public = false
private = true, public = true
private = false, public = false <- this is the stupid case

tiramisu24: wait why am i testing public. didnt change anything with regards to that?

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
tiramisu24: `    public GetObjectsOutput getObjects(final GetObjectsInput params, final String user)
            throws Exception {
        final Set<Integer> accessGroupIDs;
        final long t1 = System.currentTimeMillis();
        if (user != null) {
            accessGroupIDs = new HashSet<>(accessGroupProvider.findAccessGroupIds(user));
        } else {
            accessGroupIDs = Collections.emptySet();
        }
        final Set<GUID> guids = new LinkedHashSet<>();
        for (final String guid : params.getGuids()) {
            final GUID g = new GUID(guid);

            //TODO DP this is a quick fix for now, doesn't take data palettes into account
            if (accessGroupIDs.contains(g.getAccessGroupId())) {
                // don't throw an error, just don't return data
                guids.add(g);
            }
        }

        final kbasesearchengine.search.PostProcessing postProcessing =
                toSearch(params.getPostProcessing());
        final List<kbasesearchengine.search.ObjectData> objs = indexingStorage.getObjectsByIds(
                guids, postProcessing);
        final GetObjectsOutput ret = new GetObjectsOutput().withObjects(objs.stream()
                .map(this::fromSearch).collect(Collectors.toList()));
        ret.withSearchTime(System.currentTimeMillis() - t1);
        return ret;
    }`

I am looking at the getObjects in SearchMethods. It  looks like it was intentionally made that an unauth user can't get objects regardless of its public since its filtering the guids before calling elasticsearch. where is the get_objects match filter??

MrCreosote: Right, because before this PR all users were authed. Now unauthed users are allowed, that behavior is going to cause unauthed users to never get data, IIUC.

> where is the get_objects match filter??

https://github.com/kbase/KBaseSearchEngine/blob/master/KBaseSearchEngine.spec#L303

Don't worry about that for now though. Take it out in another PR.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
MrCreosote: public did change: https://github.com/kbase/KBaseSearchEngine/pull/260/files#diff-f58b37782d687c1180cb73e938fcec71R127

tiramisu24:
@MrCreosote can you review

RELEASE_NOTES.md
```diff
@@ -1,6 +1,11 @@
 Search Service MKII release notes
 =================================
 
+0.2.1
```
MrCreosote: Same # as previous version

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
```diff
@@ -85,6 +85,10 @@ private void quietLoggers() {
         ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME))
                 .setLevel(Level.INFO);
     }
+    private String getUserName(final AuthToken authPart){
```
MrCreosote: blank line

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -282,10 +283,13 @@ public SearchObjectsOutput searchObjects(SearchObjectsInput params, String user)
     @Override
     public GetObjectsOutput getObjects(final GetObjectsInput params, final String user)
             throws Exception {
-
+        final Set<Integer> accessGroupIDs;
         final long t1 = System.currentTimeMillis();
-        final Set<Integer> accessGroupIDs =
-                new HashSet<>(accessGroupProvider.findAccessGroupIds(user));
+        if (user != null) {
```
MrCreosote: User cannot be null, dead code

Should probably have a null check 

Note there's still a bug here in that authed users can't get access to public data unless it's in a workspace to which they have explicit read perms, but that's outside the scope of this PR

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: final args

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
                 is(expected.getIsObjectProperty() == null ? 1L : expected.getIsObjectProperty()));
     }
 
+    private SearchInterface setUpTestAccessFilter () throws Exception{
```
MrCreosote: extra space

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+                        .withPublic(true)))
+                .thenReturn(publicData);
+
+        when(is.searchTypes( kbasesearchengine.search.MatchFilter.getBuilder().build(),
+                new kbasesearchengine.search.AccessFilter()
+                        .withAccessGroups(set())
+                        .withPublic(true)))
+                .thenReturn(publicData);
+
+        //should not happen
```
MrCreosote: If this shouldn't happen, either 1) throw a runtime exception with .thenThrow() or 2) Mockito.verify() that it was never called.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+        assertThat("should show only public", res2.getTypeToCount(), is(publicData));
+
+        final SearchTypesOutput res3 = sm.searchTypes(new SearchTypesInput()
+                        .withAccessFilter(new AccessFilter().withWithPrivate(1L).withWithPublic(0L))
+                        .withMatchFilter(new MatchFilter()),
+                "auser");
+        assertThat("should show only private", res3.getTypeToCount(), is(privateData));
+
+
+        final SearchTypesOutput res4 = sm.searchTypes(new SearchTypesInput()
```
MrCreosote: This is a weird case, as mentioned previously. I sort of think it should throw an exception because it's guaranteed to return no data, so what's the point? 

That can wait for another PR if we go that way though.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+        allData.put("test", 1);
+        allData.put("notdefault", 1);
+
+        when(agp.findAccessGroupIds("auser"))
+                .thenReturn(adminList);
+
+        when(is.searchTypes( kbasesearchengine.search.MatchFilter.getBuilder().build(),
+                new kbasesearchengine.search.AccessFilter()
+                        .withAccessGroups(new LinkedHashSet(adminList))
+                        .withAdmin(true)
```
MrCreosote: why is admin true for these tests?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+                        .withMatchFilter(new MatchFilter()),
+                "auser");
+        assertThat("should show only private", res4.getTypeToCount(), is(new HashMap<>()));
+
+    }
+
+    @Test
+    public void testAccessFilterUnauthorizedUser() throws Exception{
+        SearchInterface sm = setUpTestAccessFilter();
+        final Map<String, Long> publicData = new HashMap<>();
```
MrCreosote: For future reference, this is easier:

`final Map<String, Long> publicData = ImmutableMap.of("notdefault", 1L);`

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +673,175 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+        final SearchTypesOutput res2 = sm.searchTypes(new SearchTypesInput()
+                        .withAccessFilter(new AccessFilter().withWithPrivate(0L).withWithPublic(1L))
+                        .withMatchFilter(new MatchFilter()),
+                null);
+        assertThat("incorrect counts", res2.getTypeToCount(), is(publicData));
+
+
+    }
+
+    @Test
```
MrCreosote: is this another leftover test that should be deleted?

MrCreosote (changes requested):

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
tiramisu24: what?

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: b/c I stubbed auser as an admin

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: yes

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
MrCreosote: But why did you do that?

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
MrCreosote: **blank line**

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: b/c at one point I thought i needed to test that too. I guess i can remove it

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
tiramisu24: so remove this and add the mockito verify to the top condition

tiramisu24: so if I am throwing an exception in the above stubbed case, should I have this in a try catch and see if it got the right exception or just remove this case

MrCreosote: I'm not sure - I assumed the stubbed case above was impossible. If it's possible that case could occur (or will occur in your tests) then throwing an exception isn't the right thing to do. 

IIUC, the stubbed case is when a user essentially makes a call that says "don't do a search", right? If so, that's currently allowed by the api, and I think goes to elasticsearch which will return, effectively, an empty list.

Ugh, you know what, it'd be way simpler if we just made the method throw an IllegalArgumentException if public and private are both false and make a quick test case for it. Then we don't have to think about this, assuming I'm understanding the issue.

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
tiramisu24: what do you mean blank line

tiramisu24:
@MrCreosote hopefully last time

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
MrCreosote: Add a blank line here, between the functions

lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
```diff
@@ -760,7 +760,7 @@ Retrieve objects by their GUIDs.
 {
     my($self, @args) = @_;
 
-# Authentication: required
+# Authentication: requred
```
MrCreosote: This is misspelled, which is odd since it's a generated file. Did you edit the generated files manually? If so just recompile

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -122,15 +122,23 @@ private static Integer toInteger(Long value) {
     }
 
     private kbasesearchengine.search.AccessFilter toSearch(AccessFilter af, String user)
-            throws IOException {
+            throws IOException, IllegalArgumentException {
         List<Integer> accessGroupIds;
-        if (toBool(af.getWithPrivate(), true)) {
+        final boolean withPublic = (user == null) ? true : toBool(af.getWithPublic());
+
+        boolean test = !toBool(af.getWithPrivate(), true);
```
MrCreosote: Unused

lib/src/kbasesearchengine/KBaseSearchEngineServer.java
MrCreosote: blank line

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -122,15 +122,23 @@ private static Integer toInteger(Long value) {
...
+        if(!withPublic &&  !toBool(af.getWithPrivate(), true)){
```
MrCreosote: space after if, extra space before !, missing space before {

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -122,15 +122,23 @@ private static Integer toInteger(Long value) {
...
+            throw new IllegalArgumentException("should have either public or private data");
```
MrCreosote: This message would be really hard to understand for someone using the client.

test/src/kbasesearchengine/test/integration/SearchAPIIntegrationTest.java
MrCreosote: final args

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +672,120 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+                "auser");
+        assertThat("should show only public", res2.getTypeToCount(), is(publicData));
+
+        final SearchTypesOutput res3 = sm.searchTypes(new SearchTypesInput()
+                        .withAccessFilter(new AccessFilter().withWithPrivate(1L).withWithPublic(0L))
+                        .withMatchFilter(new MatchFilter()),
+                "auser");
+        assertThat("should show only private", res3.getTypeToCount(), is(privateData));
+
+        //should not happen
```
MrCreosote: This should be a separate test

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +672,120 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+        try{
```
MrCreosote: lots of spacing issues for the rest of this function

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +672,120 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+            final SearchTypesOutput res4 = sm.searchTypes(new SearchTypesInput()
+                            .withAccessFilter(new AccessFilter().withWithPrivate(0L).withWithPublic(0L))
+                            .withMatchFilter(new MatchFilter()),
+                    "auser");
+        }catch(IllegalArgumentException e){
+            assertThat("should throw exception if private and public set to false"
+                    , e.getMessage(), is("should have either public or private data"));
+        }catch(Exception e){
```
MrCreosote: You don't want to do this, this will mask other exceptions that might occur.

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +672,120 @@ private void compare(final SortingRule got, final SortingRule expected) {
...
+
+        //should not happen
+        try{
+            final SearchTypesOutput res4 = sm.searchTypes(new SearchTypesInput()
+                            .withAccessFilter(new AccessFilter().withWithPrivate(0L).withWithPublic(0L))
+                            .withMatchFilter(new MatchFilter()),
+                    "auser");
+        }catch(IllegalArgumentException e){
+            assertThat("should throw exception if private and public set to false"
+                    , e.getMessage(), is("should have either public or private data"));
```
MrCreosote: comma on previous line

test/src/kbasesearchengine/test/main/SearchMethodsTest.java
```diff
@@ -669,4 +672,120 @@ private void compare(final SortingRule got, final SortingRule expected) {
                 is(expected.getIsObjectProperty() == null ? 1L : expected.getIsObjectProperty()));
     }
 
+    private SearchInterface setUpTestAccessFilter () throws Exception{
```
MrCreosote: extra space

RELEASE_NOTES.md
```diff
@@ -1,6 +1,11 @@
 Search Service MKII release notes
 =================================
 
+0.2.2
+-----
+* Changed search_objects and search_types to optionally require auth. Unauthorized calls will only
+see results from public data.
+
```
MrCreosote: Note that now selecting public = private = False will throw an exception

MrCreosote (changes requested):

tiramisu24:
@MrCreosote I think it can be merged

MrCreosote:
? Most of the issues are unresolved.

MrCreosote:
@tiramisu24 
Merging this also merged https://github.com/kbase/KBaseSearchEngine/pull/236. Is that what you expected? That surprised the heck out of me. Can you check the code is what you expect?

PR #261, opened 2018-07-17
Sct 1144 DELETE_ALL_VERSIONS event

tiramisu24:
Bug fixes:
-skip check ws info and change events for DELETE_ALL_VERSION event
-change term search to prefix search for updating deleted objects


tiramisu24:
@arfathpasha @MrCreosote can one of you review?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -840,6 +838,11 @@ public StatusEvent updateObjectEvent(final StatusEvent ev)
             return ev;
         }
 
+        //should not need to update object event if event is to delete
+        if(ev.getEventType().name().equals("DELETE_ALL_VERSIONS")){
```
MrCreosote, line 842: I've realized this introduces a bug - we should not be relying on the object state (which here is stored in the event name itself) since events may be run out of order - for example, if an event fails and ops reruns all failed events later. I'll add a ticket.

Pinging @scanon fyi

tiramisu24: What are all the failure messages for if an object is delete?

MrCreosote: The workspace failure messages? I honestly don't know, this is one of the areas where the workspace could use some TLC.

tiramisu24: So the current EventHandler just checks if objects are deleted if events come out of sync. Based on what you said earlier, the one message check on cut it. Is there a good way to tell if an object is deleted or should I start scouring the workspace for failure messages?

MrCreosote: I would check the same way the current code checks. There's code in there, that given an event, will convert it into a deleted event if the object is deleted. You could probably reuse it to convert a deletion event to an undelete event if the object is not deleted.

tiramisu24: The current code does the error parsing and does not catch all the errors

MrCreosote: How do you know that?

PR #263, opened 2018-08-01
Sct 1144

tiramisu24:
-continuation of sct 1144
-added total objects looked at 
-loops to get enough objects requested in pagination

### DIFF
--- a/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
+++ b/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
@@ -88,7 +88,7 @@ public class NarrativeInfoDecorator implements SearchInterface {
             throws Exception {
         SearchObjectsOutput searchObjsOutput = getEmptySearchObjectsOutput();
         long count = (params.getPagination() == null) ? 50 : params.getPagination().getCount();
-        while(searchObjsOutput == null || searchObjsOutput.getObjects().size() < count){
+        while(searchObjsOutput.getObjects().size() < count){
             final SearchObjectsOutput searchRes = reSearchObjects(params, user, searchObjsOutput);
 
             if (params.getPostProcessing() != null && params.getPostProcessing().getAddNarrativeInfo() != null
@@ -200,7 +200,9 @@ public class NarrativeInfoDecorator implements SearchInterface {
                                                 .withMatchFilter(params.getMatchFilter())
                                                 .withAccessFilter(params.getAccessFilter())
                                                 .withPagination(newPag);
-            res = combineWithOtherSearchObjectsOuput(res, searchInterface.searchObjects(newParams, user), 0);
+
+            SearchObjectsOutput temp =  searchInterface.searchObjects(newParams, user);
+            res = combineWithOtherSearchObjectsOuput(res, temp, 0);
 
         }
         return res;
--- a/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
+++ b/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
@@ -277,9 +277,11 @@ public class NarrativeInfoDecorator implements SearchInterface {
 
                 if(curTotal >= targetTotal){
                     //remove extra results
-                    do{
+                    while(iter.hasNext()){
                         iter.remove();
-                    }while(iter.hasNext());
+                        iter.next();
+                    };
+                    iter.remove();
                     break;
                 }else if(deletedWsIdsSet.contains(workspaceId)){
                     iter.remove();
--- a/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
+++ b/lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
@@ -80,15 +80,25 @@ public class NarrativeInfoDecorator implements SearchInterface {
     @Override
     public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final String user)
             throws Exception {
-        SearchObjectsOutput searchObjsOutput = searchInterface.searchObjects(params, user);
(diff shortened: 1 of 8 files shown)


lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -80,18 +86,126 @@ public SearchTypesOutput searchTypes(final SearchTypesInput params, final String
...
-        SearchObjectsOutput searchObjsOutput = searchInterface.searchObjects(params, user);
-        if (params.getPostProcessing() != null) {
-            if (params.getPostProcessing().getAddNarrativeInfo() != null &&
-                    params.getPostProcessing().getAddNarrativeInfo() == 1) {
-                searchObjsOutput = searchObjsOutput.withAccessGroupNarrativeInfo(addNarrativeInfo(
-                        searchObjsOutput.getObjects(),
-                        searchObjsOutput.getAccessGroupNarrativeInfo()));
+        SearchObjectsOutput searchObjsOutput = getEmptySearchObjectsOutput();
+        long count = (params.getPagination() == null) ? 50 : params.getPagination().getCount();
+        while(searchObjsOutput == null || searchObjsOutput.getObjects().size() < count){
```
arfathpasha: searchObjsOutput can never be null here right?

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +377,66 @@ public void listTypes() throws Exception {
...
+        final NarrativeInfoDecorator nid = setUpSearchObjectsWithDeletedWs();
+        final Pagination pag = new Pagination().withCount(3L).withStart(0L);
+        final SearchObjectsInput dummyInput = new SearchObjectsInput()
+                .withPostProcessing(new PostProcessing().withAddNarrativeInfo(1L))
+                .withPagination(pag);
+
+        final ArrayList<ObjectData> expectedObjs = new ArrayList<>();
+        expectedObjs.add(new ObjectData().withGuid("WS:4/1/7").withCreator("user"));
+        expectedObjs.add(new ObjectData().withGuid("WS:4/2/1").withCreator("user"));
+        expectedObjs.add(new ObjectData().withGuid("WS:4/1/7").withCreator("user"));
```
arfathpasha, line 391: Shouldn't there be a WS:3/5/6 because setUpSearchObjectsWithDeletedWs() has this as one of the three guids?

I don't follow how this test is passing.

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +377,66 @@ public void listTypes() throws Exception {
...
+
+        final SearchObjectsOutput res = nid.searchObjects(dummyInput, "user");
+
+        compareSearchObjectOutputRes(res.getObjects(), expectedObjs);
+        assertThat("incorrect number of hits looked at", res.getTotal(), is(6L));
```
arfathpasha: Total should be 3 since there are only three objects being returned in res?

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
arfathpasha, line 391: It would be good to have a test where the requested count is not fulfilled (because of deleted workspaces, but simulated of course) and subsequent calls to reSearchObjects() and combineWithOtherSearchObjectOutput() are made in the searchObjects() while loop.

arfathpasha (changes requested):

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
tiramisu24: changing total to be the total number of objects looked at so the ones removed won't mess up pagination

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
tiramisu24: anything from WS:/3 should be removed as it is shown to be deleted. Currently they are returned as null, which is a problem as the front end sometimes gets pages of deleted workspaces.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: ops

MrCreosote:
A title & body that gives some description of what's going on here would help

MrCreosote:
This seems to be the hub of the PR:
> Keep call elasticsearch until enough objects are collected to satisfy pagination.

I don't think we should do this. The only way data can be removed from the results returned from ES is when the search for data occurs between the time the data deletion occurs and the event is processed by the indexing system. That's going to be a very rare case if the indexer is working properly (which we know it isn't currently). Rather than adding complexity to an already overly complex and difficult to maintain and change system, we should figure out what is preventing the deletion event from being processed in a timely manner and fix that, whether it's bugfixes, deploying more indexer workers, or whatever.

tiramisu24:
So the first part of SCT-1144 was to fix the issues with deleted objects. This pr is actually just adding the logic of re-calling elasticsearch if there is not enough objects to fill pagination.

MrCreosote:
If the deleted object issue is fixed, then there's no need for this PR as I understand it.

tiramisu24:
Initially I was told there was a consensus on wanting to fetch new results from elastic. This just got revised to setting the search results to null (when its deleted/not accessible). Going to close this pr and open a new one

PR #264, opened 2018-08-06
SCT-1203

tiramisu24:
-fix bug that sets withPrivate as true as default
-add tests for searchObjects

### DIFF
--- a/lib/src/kbasesearchengine/main/SearchMethods.java
+++ b/lib/src/kbasesearchengine/main/SearchMethods.java
@@ -54,11 +54,13 @@ public class SearchMethods implements SearchInterface {
         this.typeStorage = typeStorage;
         this.indexingStorage = indexingStorage;
     }
-    
-    private static boolean toBool(Long value) {
-        return value != null && value == 1L;
-    }
 
+    /**
+     * Changes Long to Boolean. 1L = true. Anything else is False. Accepts default value
+     * @param value       Intended value
+     * @param defaultRet  Default boolean
+     * @return
+     */
     private static boolean toBool(Long value, boolean defaultRet) {
         if (value == null) {
             return defaultRet;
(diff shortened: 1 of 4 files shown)


lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -125,16 +125,13 @@ private static Integer toInteger(Long value) {
             throws IOException, IllegalArgumentException {
         List<Integer> accessGroupIds;
         final boolean withPublic = (user == null) ? true : toBool(af.getWithPublic());
-
-        if (!withPublic && !toBool(af.getWithPrivate(), true)) {
+        final boolean withPrivate = (user == null) ? false : toBool(af.getWithPrivate(), true);
```
bio-boris: Intellij suggests changing this to
```
        final boolean withPublic = (user == null) || toBool(af.getWithPublic());
        final boolean withPrivate =  (user != null) && toBool(af.getWithPublic(),true);
```

lib/src/kbasesearchengine/main/SearchMethods.java
```diff
@@ -125,16 +125,13 @@ private static Integer toInteger(Long value) {
             throws IOException, IllegalArgumentException {
```
bio-boris, line 136: Is there documentation for these functions anywhere? If not, do you think it would be useful to add a docstring before the function definition to help understand what this function is doing?

lib/src/kbasesearchengine/main/SearchMethods.java
tiramisu24: With the suggested changes, is it clear to you that for an unauth user (user == null) withPublic and withPrivate is hardcoded to true and false? If so I can add change it.

lib/src/kbasesearchengine/main/SearchMethods.java
tiramisu24: There is a todo for javadoc in the interface. Do we need it for private methods?

bio-boris: I think the comment goes at the top of the method as in

https://en.wikipedia.org/wiki/Javadoc#Examples

bio-boris: @tiramisu24
Also the format for variables are missing a description
`@param  variable Description`

  Also I had a question, why does withPublic not have a default boolean value?

bio-boris (commented):
It looks like its working as intended, not sure we need to actually remove the toBool function though. Sorry, I thought you meant changing to 
"""
     final boolean withPublic = (user == null) ? true : toBool(af.getWithPublic(), false);
       final boolean withPrivate = (user == null) ? false : toBool(af.getWithPrivate(), true);
"""
not by actually removing the function,

bio-boris (approved):

PR #265, opened 2018-08-08
fix incorrect workspace info field used for public permission

eapearson:
fixes [SCT-1203], private data exposed in unauthenticated search.
This was due to a reference to the wrong workspace info field, getE6, rather than getE7.
The public permissions is stored in position 6 (0-based) which translates to getE7 (1-based).

I have not definitively traced down the introduction of the bug, don't know if it is important. It is not in production, just CI. It looks like this bug came in around here:
https://github.com/kbase/KBaseSearchEngine/commit/78c385a68d3feb392147d67b591bf8a410039654#diff-2ae08fe926f1612b2ee2394ef98f6b26R818

There was quite a bit of churn in this file, with the correct code disappearing, then appearing with this error, then disappearing, then the error recurs, and all the while the framing of the code changes quite a bit. I suspect dueling codebases?

### DIFF
--- a/test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
+++ b/test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
@@ -1,3 +1,4 @@
+
 package kbasesearchengine.test.events.handler;
 
 import static org.hamcrest.CoreMatchers.is;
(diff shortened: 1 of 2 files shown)


test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
```diff
@@ -1472,4 +1473,123 @@ public void newVersionEventWithOverwriteExistingData() throws Exception {
         Assert.assertEquals(expectedEvent, updatedEvent);
     }
 
+    /*
+    Test that a NEW_VERSION object event with the isPublic flag set to false will be updated
```
tiramisu24: this sentence keeps going and not sure what its saying

test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
```diff
@@ -1472,4 +1473,123 @@ public void newVersionEventWithOverwriteExistingData() throws Exception {
...
+                Instant.ofEpochMilli(20000),
+                StatusEventType.NEW_VERSION)
+                .withNullableOverwriteExistingData(true)  // overwrite existing data
+                .withNullableisPublic(Boolean.TRUE)
+                .withNullableAccessGroupID(3)
+                .withNullableObjectID("1")
+                .withNullableVersion(1)
+                .build();
+
+        Assert.assertNotSame("expected same event object after update", event, updatedEvent);
```
tiramisu24: comment should be "expected different event object after update"

test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
```diff
@@ -1472,4 +1473,123 @@ public void newVersionEventWithOverwriteExistingData() throws Exception {
...
+                .withNullableObjectID("1")
+                .withNullableVersion(1)
+                .build();
+
+        List<Tuple11<Long, String, String, String,
+                Long, String, Long, String, String, Long, Map<String, String>>> objList = new ArrayList<Tuple11<Long, String, String, String, Long, String, Long, String, String, Long, Map<String,String>>>();
+        objList.add(objTuple(1L, null, "sometype", "date", 1L, "copier",
+                3L, "wsname", "checksum", 44, Collections.emptyMap()));
+
+        when(wscli.administer(any()))
```
tiramisu24, line 1511: this could be a little unclear later on? maybe add the method in the comments too?

eapearson: I'm not sure what you mean here.

test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
eapearson: I've actually removed this assertion from the upcoming change to this file - it is rather pointless. It is actually copied from the test above. It seems to me too general an assertion to make - yes they are different but the assertion below it which indicates that the event should have been updated in a certain way is both more specific and also covers the case of it being different.

test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
eapearson: This test was copied from an earlier test. I'm not too familiar with mockito, but I did spend a few minutes reading up on it so that I could understand how this mechanism works. If you are referring to the comment above the mockito stuff explaining how it is supposed to work, I've removed that.

PR #266, opened 2018-08-10
fix public permissions test [SCT-1203]

eapearson:
- the string equalty check unwisely used ==
- it passed the tests because the mock workspace api uses static strings
- it was also more complex than necessary
- it also used a double-negative

In the affected code, the "isPublic" flag was determined by comparing the current public permission for a workspace to the string "n", using the == operator. 

Now it compares isPublic to the read-only permission "r" using the equals method.

The previous "fix" failed to work in real life but passed the associated test because the test uses a mock workspace client which is populated with literal values. In real usage, the workspace client fetches the values from the database, and the values are objects not literals. In the associated test the public permission value is now provided as a string object, which indeed caused the test to fail if the code was not fixed to use equals over ==.

I also searched the codebase for other usages of string comparisons using == and could find none.

### DIFF
--- a/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
+++ b/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
@@ -821,7 +821,7 @@ public class WorkspaceEventHandler implements EventHandler, WorkspaceInfoProvide
             final Long wsId = Long.valueOf(ev.getAccessGroupId().get()).longValue();
             final String isPublic = getWorkspaceInfo(wsId).getE7();
 
-            return (isPublic == "n") ? false: true;
+            return isPublic.equals("n") ? false: true;
 
         } catch (IOException ex) {
             throw handleException(ex);
--- a/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
+++ b/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
@@ -821,7 +821,12 @@ public class WorkspaceEventHandler implements EventHandler, WorkspaceInfoProvide
             final Long wsId = Long.valueOf(ev.getAccessGroupId().get()).longValue();
             final String isPublic = getWorkspaceInfo(wsId).getE7();
 
-            return isPublic.equals("n") ? false: true;
+            // The workspace info array element 6 (tuple element 7) is the public permission, 
+            // and represents the access permitted to the workspace without authentication,
+            // i.e. without an identified kbase user.
+            // It is either "n" indicating not publicly accessible, or "r" indicating
+            // that it is publicly readable.
+            return isPublic.equals("r");
 
         } catch (IOException ex) {
             throw handleException(ex);
--- a/test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
+++ b/test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
@@ -1514,7 +1514,7 @@ public class WorkspaceEventHandlerTest {
                 // for name check
                 .thenReturn(new UObject(new GetObjectInfo3Results().withInfos(objList)))
                 // for isPublic check
-                .thenReturn(new UObject(wsTuple(3L, "wsname", "username", "date", 7, "n", "r",
+                .thenReturn(new UObject(wsTuple(3L, "wsname", "username", "date", 7, "n", new String("r"),
                         "unlocked", Collections.emptyMap())));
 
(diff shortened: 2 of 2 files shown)


MrCreosote:
Just FYI, there's no such thing as 'static' or 'literal' strings in compiled Java. When the compiler sees a string literal it turns it into a string object, so you can do things like `if ("n".equals(var)) {...`

What might be happening is [string interning](https://stackoverflow.com/questions/10578984/what-is-java-string-interning).

eapearson:
Thanks for the elucidation :) I knew that string literal usage was not resulting in a primitive type, but somehow two string objects which are the same, so now I know, interning. Consider my brain tainted by javascript, in which a string literal is a string type (comparable by equality), and a string object is an object type which happens to have the String prototype (not comparable by equality unless they happen to be the same object). Works similarly, though, at least in the simple cases.

eapearson:
BTW would you mind merging this if you see fit, @MrCreosote? Diane is out today and I want to close this, if possible, while technically still in this sprint.

eapearson:
Hi @tiramisu24 would mind having a look at this and merge if it looks okay 2u?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -821,7 +821,12 @@ private Boolean getLatestIsPublic(StatusEvent ev)
             final Long wsId = Long.valueOf(ev.getAccessGroupId().get()).longValue();
             final String isPublic = getWorkspaceInfo(wsId).getE7();
 
-            return (isPublic == "n") ? false: true;
```
tiramisu24: how is different

test/src/kbasesearchengine/test/events/handler/WorkspaceEventHandlerTest.java
```diff
@@ -1514,7 +1514,7 @@ public void newVersionPublicFlagUpdatedFromWorkspaceCase1() throws Exception {
                 // for name check
                 .thenReturn(new UObject(new GetObjectInfo3Results().withInfos(objList)))
                 // for isPublic check
-                .thenReturn(new UObject(wsTuple(3L, "wsname", "username", "date", 7, "n", "r",
+                .thenReturn(new UObject(wsTuple(3L, "wsname", "username", "date", 7, "n", new String("r"),
```
tiramisu24, line 1517: why new String

eapearson: This is tied to the fix to replace == for string comparison with equals. When using == two strings are only equal if they are the same string object or are interned strings (e.g. derived form the same string literal or using new String("foo").intern()). When you say "foo" the string literals are converted to "interned" string objects -- a special pool of string objects in the JVM for which there is only one object for each unique string. Each of the "foo"s in "foo" == "foo" resolves to the same interned string object, and so it evaluates to true. But when you compare non-interned strings, even if the string is the same, == gives you false, since they are different objects. In all cases equals works as expected.

The problem with using string literals in the test in this case is that in the actual workspace api, the string values returned are always (at present) non-interned string objects. Thus, if you get the public permission value from a workspace api call, and compare it to some other value  some string value with == it will always be false. However, the workspace mock returns exactly what you give to the "thenReturn" method, which in this case was a string literal / interned string object.

The WorkspaceEventHandler.getLatestIsPublic, which determines the current state of the public permission for the workspace, uses a string literal "n" for comparison.

So the pre-fixed code in WorkspaceEventHandler.getLatestIsPublic used == and would succeed in the test, but fail in real life when a real workspace call was made. 

So, replacing == with equals fixed that problem.

On the other hand, using "n" for the mock is wrong because the non-mocked api returns a non-interned string.

So, many times more words than required, the usage of new String("r") is more correct than "r" for the current behavior of the workspace api. (Though admittedly the workspace api could return interned strings at some point in the future.) However, after replacing == with equals in WorkspaceEventHandler.getLatestIsPublic it doesn't make a difference.

In my opinion, the mock should accurately reflect the precise nature of the real api in this respect, because otherwise it can lead test results to differ from real-life results in subtle ways. Imagine someone in the future decides to replace equals with == in getLatestIsPublic, and feels proud because it is more efficient and the tests still pass. It would fail in a real deployment.

In this case I only "fixed" the one field of the get_workspace_info simulated result, since it is the only one used in this test which would be affected.

eapearson: Something like that, I'm no java pro, though.

PR #268, opened 2018-08-20
add test

tiramisu24:
-continuation of SCT 1096
-change index to * when searching for latest object
-add test for an object with subobject

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -595,7 +595,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
     private Integer loadLastVersion(GUID parentGUID,
             Integer processedVersion) throws IOException {
 
-        final String reqIndexName = "*";
+        final String reqIndexName = getAnyIndexPattern();
 
         String prefix = toGUIDPrefix(parentGUID);
 
@@ -653,7 +653,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
     private int updateLastVersionsInData(GUID parentGUID,
             int lastVersion) throws IOException, IndexingConflictException {
 
-        final String indexName = "*";
+        final String indexName = getAnyIndexPattern();
 
         // query = {"bool": {"filter": [{"term": {"prefix": prefix}}]}}
         Map<String, Object> query = ImmutableMap.of("bool",
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -350,7 +350,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
         File tempFile = File.createTempFile("es_bulk_", ".json", tempDir);
         try {
             PrintWriter pw = new PrintWriter(tempFile);
-            int lastVersion = loadLastVersion(indexName, pguid, pguid.getVersion());
+            int lastVersion = loadLastVersion(pguid, pguid.getVersion());
             final String esParentId = checkParentDoc(indexName, new LinkedHashSet<>(
                     Arrays.asList(pguid)), isPublic, lastVersion).get(pguid);
             if (idToObjCopy.isEmpty()) {
@@ -378,7 +378,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
 
             verify(response);
 
-            updateLastVersionsInData(indexName, pguid, lastVersion);
+            updateLastVersionsInData(pguid, lastVersion);
         } finally {
             tempFile.delete();
         }
@@ -592,10 +592,10 @@ public class ElasticIndexingStorage implements IndexingStorage {
                 Collectors.toMap(Function.identity(), guid -> map.containsKey(guid))));
     }
 
-    private Integer loadLastVersion(String reqIndexName, GUID parentGUID, 
+    private Integer loadLastVersion(GUID parentGUID,
             Integer processedVersion) throws IOException {
 
-        reqIndexName = getGeneralIndexName(reqIndexName);
+        final String reqIndexName = "*";
 
         String prefix = toGUIDPrefix(parentGUID);
 
@@ -644,17 +644,16 @@ public class ElasticIndexingStorage implements IndexingStorage {
      * The search in step 2 requires the refresh on every bulk indexing and update operation,
      * else we get the version conflict exceptions.
      *
-      * @param indexName
      * @param parentGUID
      * @param lastVersion
      * @return
      * @throws IOException
      * @throws IndexingConflictException
      */
-    private int updateLastVersionsInData(String indexName, GUID parentGUID,
+    private int updateLastVersionsInData(GUID parentGUID,
             int lastVersion) throws IOException, IndexingConflictException {
 
-        indexName = getGeneralIndexName(indexName);
+        final String indexName = "*";
 
         // query = {"bool": {"filter": [{"term": {"prefix": prefix}}]}}
         Map<String, Object> query = ImmutableMap.of("bool",
@@ -681,6 +680,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
         String urlPath = "/" + indexName + "/" + getDataTableName() + "/_update_by_query?refresh=wait_for";
         Response resp = makeRequest("POST", urlPath, doc);
         @SuppressWarnings("unchecked")
+
         Map<String, Object> data = UObject.getMapper().readValue(
                 resp.getEntity().getContent(), Map.class);
         return (Integer)data.get("updated");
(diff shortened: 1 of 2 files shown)


test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1815,4 +1815,45 @@ public void testDeleteObjectWithSubobject() throws Exception {
...
+                Collections.emptyList(),
+                MatchFilter.getBuilder().withNullableFullTextInAll("imaprettypony").build(),
+                null,
+                AccessFilter.create().withAccessGroups(2000));
+        assertThat("incorrect objects found", res,
+                is(set(new GUID("WS:2000/1/2"), new GUID("WS:2000/1/2:sub/yay"))));
+
+    }
+
+    public void indexObject(final String pguid, final String guid, final int ver)throws Exception{
```
MrCreosote: Should make it clear that the ver here is the search type version, not the source object version. That confused me for a bit.

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1815,4 +1815,45 @@ public void testDeleteObjectWithSubobject() throws Exception {
 
         assertThat("objects not deleted", res2, is(set()));
     }
+    @Test
```
MrCreosote, line 1818: If this is the test we discussed previously, the case I was worried about was this:

1) Index an object with subobjects (so records in 2 indexes would be updated)
2) Index a new version of the object that either has a different or missing SearchType for the main object or the subobjects or both (imagine saving over a genome with reads, for example)
3) Would that cause the records of the older version of the objects to have `islast` set to false?

MrCreosote (changes requested):

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: Why would the SearchType be different? Wouldn't it just be the search type versions?

MrCreosote: Because the storage object type is different. A genome will have a different (set of) search type than reads. Reads probably won't have a sub object type at all, while a genome will have GenomeFeatures.

MrCreosote: For example:
```python
In [11]: ws.create_workspace({'workspace': 'dztest'})
Out[11]: 
[35854,
 u'dztest',
 u'gaprice',
 u'2018-08-22T20:41:31+0000',
 0,
 u'a',
 u'n',
 u'unlocked',
 {}]

In [12]: ws.save_objects({'workspace': 'dztest', 'objects': [{'type': 'Empty.AType-1.0', 'data': {'foa': 'bar'}, 'name': 'dz'}]})
Out[12]: 
[[1,
  u'dz',
  u'Empty.AType-1.0',
  u'2018-08-22T20:41:37+0000',
  1,
  u'gaprice',
  35854,
  u'dztest',
  u'c68986e24d15817eac2b6e98b1dad6a7',
  13,
  {}]]

In [13]: ws.save_objects({'workspace': 'dztest', 'objects': [{'type': 'Empty.ARef-1.0', 'data': {'ref': 'dztest/dz'}, 'name': 'dz'}]})
Out[13]: 
[[1,
  u'dz',
  u'Empty.ARef-1.0',
  u'2018-08-22T20:41:42+0000',
  2,
  u'gaprice',
  35854,
  u'dztest',
  u'ba99531e85ba40b05ee9a09ffa31dd62',
  19,
  {}]]
```

These are obviously just test types, but they demonstrate that a new version of an object could have a completely different workspace type, and therefore a completely different set of search types.

tiramisu24: wait why they considered different versions of an object if their types are different? Is the versions solely based on the name of the object??

tiramisu24: so if I had an e coli genome and I named it to rhodobacter.reads and then added the actual rhodobacter reads, the e coli genome would be overridden by the reads?

MrCreosote: > Is the versions solely based on the name of the object??

Yep

MrCreosote: To both questions

MrCreosote: https://ci.kbase.us/services/ws/docs/fundamentals.html

tiramisu24: In that case, your suspicions are correct. If the two objects on different search types, the first ver object will not have isLast set to false.

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -592,10 +592,10 @@ public void flushIndexing(final ObjectTypeParsingRules rule) throws IOException
                 Collectors.toMap(Function.identity(), guid -> map.containsKey(guid))));
     }
 
-    private Integer loadLastVersion(String reqIndexName, GUID parentGUID, 
+    private Integer loadLastVersion(GUID parentGUID,
             Integer processedVersion) throws IOException {
 
-        reqIndexName = getGeneralIndexName(reqIndexName);
+        final String reqIndexName = "*";
```
MrCreosote: You don't want to do this as it will search outside the namespace.

MrCreosote (changes requested):

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: Isn't this what we want? How else would you know what is the last version of something?

MrCreosote: No, we don't want to search outside of the application namespace for sure. There obviously won't be any relevant records other than inside the namespace.

tiramisu24: By namespace, do you mean the index name? I am not sure what you mean by application namespace

MrCreosote: Resolved this offline

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -592,10 +592,10 @@ public void flushIndexing(final ObjectTypeParsingRules rule) throws IOException
                 Collectors.toMap(Function.identity(), guid -> map.containsKey(guid))));
     }
 
-    private Integer loadLastVersion(String reqIndexName, GUID parentGUID, 
+    private Integer loadLastVersion(GUID parentGUID,
             Integer processedVersion) throws IOException {
 
-        reqIndexName = getGeneralIndexName(reqIndexName);
+        final String reqIndexName = getAnyIndexPattern();
```
MrCreosote, line 598: These changes could come with pretty severe performance penalties. I can think of a few potential ways to alleviate these putative penalties:

1) Do an aggregation when looking up the latest version that only records the latest version and the indexes in which the object resides (as determined by the prefix). This should reduce memory use and transport time (especially for object versions with lots of subobjects).

2) Use the indexes gathered in step 1 to only perform changes on the indexes in which the object resides, instead of all of them. I'm not sure whether this will make a huge difference unless there are a lot of indexes - each index that doesn't contain the object will incur a O(log N) penalty determining it has nothing to do, but this might be negligible compared to the cost of writing the changes in the other indexes.

3) The script, IIRC, writes a new document *even if islast doesn't change*. Change the script (if possible) so that it only causes a new document to be written if is last changes. Alternatively, confirm my understanding is wrong.

Maybe we should discuss what to do (e.g. make changes, benchmark, nothing) about this with @scanon and @sychan at the standup Monday.

tiramisu24: https://www.elastic.co/guide/en/elasticsearch/reference/5.2/docs-update-by-query.html
It looks like if we set ctx.op = "noop" in the script we can skip updating the document

tiramisu24: I am not sure aggregation would make things faster as it still searches for all the results before aggregating them?

MrCreosote: > This should reduce memory use and transport time (especially for object versions with lots of subobjects).

MrCreosote: I also think the update could be simplified - right now now it touches every document. Instead do something like

```
FIND documents WHERE ver < lastestver AND guid == expectedGuid AND islast == true:
   SET islast == false
FIND documents WHERE ver == lastestver AND guid == expected GUID AND islast == false:
   SET islast == true
```

Maybe simpler ways than that as well.

tiramisu24: So this is something I am not sure about. When getting lastestver, the query searches everywhere for the expected guid, and gets the highest version from that. So when the second query (to update) fires, would elastic use the cached first query or see the second query as totally different?

MrCreosote: I don't know, but I'd expect not.

scanon:
I'm closing this since we have frozen things for some time

PR #269, opened 2018-08-20
[SCT-1144] Fix bug preventing deletion events. Remove deleted objects from results

tiramisu24:
-remove result from returned list if workspace is deleted
-adds guid to list of inaccessible guids
-Fix bug preventing delete events from going through.

### DIFF
--- a/KBaseSearchEngine.spec
+++ b/KBaseSearchEngine.spec
@@ -261,8 +261,8 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
-      list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-          not been updated in search. Returned when setting env SHOWREMOVEDGUIDS to anything. Returned in additional properties
+      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
+         objects will match input from pagination.
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -307,8 +307,7 @@ module KBaseSearchEngine {
 
     /*
       Output results of get_objects method.
-      list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-          not been updated in search. Returned when setting env SHOWREMOVEDGUIDS to anything. Returned in additional properties
+      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed.
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
--- a/KBaseSearchEngine.spec
+++ b/KBaseSearchEngine.spec
@@ -262,7 +262,7 @@ module KBaseSearchEngine {
       total - total number of found objects.
       search_time - common time in milliseconds spent.
       list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-          not been updated in search
+          not been updated in search. Returned when setting env SHOWREMOVEDGUIDS to anything. Returned in additional properties
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -282,7 +282,6 @@ module KBaseSearchEngine {
         list<ObjectData> objects;
         int total;
         int search_time;
-        list<GUID> removedGuids;
         mapping<access_group_id, narrative_info> access_group_narrative_info;
         mapping<access_group_id, access_group_info> access_groups_info;
         mapping<obj_ref, object_info> objects_info;
@@ -309,7 +308,7 @@ module KBaseSearchEngine {
     /*
       Output results of get_objects method.
       list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-          not been updated in search
+          not been updated in search. Returned when setting env SHOWREMOVEDGUIDS to anything. Returned in additional properties
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -326,7 +325,6 @@ module KBaseSearchEngine {
     typedef structure {
         list<ObjectData> objects;
         int search_time;
-        list<GUID> removedGuids;
         mapping<access_group_id, narrative_info> access_group_narrative_info;
         mapping<access_group_id, access_group_info> access_groups_info;
         mapping<obj_ref, Workspace.object_info> objects_info;
--- a/KBaseSearchEngine.spec
+++ b/KBaseSearchEngine.spec
@@ -261,6 +261,8 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
+      list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
+          not been updated in search
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -280,6 +282,7 @@ module KBaseSearchEngine {
         list<ObjectData> objects;
         int total;
         int search_time;
+        list<GUID> removedGuids;
         mapping<access_group_id, narrative_info> access_group_narrative_info;
         mapping<access_group_id, access_group_info> access_groups_info;
         mapping<obj_ref, object_info> objects_info;
@@ -305,7 +308,8 @@ module KBaseSearchEngine {
 
     /*
       Output results of get_objects method.
-      
+      list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
+          not been updated in search
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -322,6 +326,7 @@ module KBaseSearchEngine {
     typedef structure {
         list<ObjectData> objects;
         int search_time;
+        list<GUID> removedGuids;
         mapping<access_group_id, narrative_info> access_group_narrative_info;
         mapping<access_group_id, access_group_info> access_groups_info;
         mapping<obj_ref, Workspace.object_info> objects_info;
--- a/KBaseSearchEngine.spec
+++ b/KBaseSearchEngine.spec
@@ -261,8 +261,11 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
-      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
-         objects will match input from pagination.
+      list<ObjectData> objects - Results of search. In rare cases, due to time delay in updating the search
+        index from  changes in the source data, objects returned from the search
+        index may not be accessible at the data source. The API will remove
+        these objects from the returned search results which means the returned
+        results may have fewer objects than a full page.
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
@@ -307,7 +310,10 @@ module KBaseSearchEngine {
 
     /*
       Output results of get_objects method.
-      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed.
+      list<ObjectData> objects - Results of search. In rare cases, due to time delay in updating the search
+        index from  changes in the source data, objects returned from the search
+        index may not be accessible at the data source. The API will remove
+        these objects from the returned search results.
       mapping<access_group_id, narrative_info> access_group_narrative_info - information about
          the workspaces in which the objects in the results reside. This data only applies to
          workspace objects.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,11 @@
 Search Service MKII release notes
 =================================
+0.2.4
+---
+* Remove objects in deleted/inaccessible workspaces from search results. This should be a rare event that, 
+  due to time delay in updating the searchvindex from  changes in the source data,
+  objects returned from the search index may not be accessible at the data source. 
+
 0.2.3
 ----
 * Upgrade ES from 5.5 to 5.6.11
--- a/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
+++ b/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
@@ -313,8 +313,8 @@ SearchObjectsOutput is a reference to a hash where the following keys are define
 	sorting_rules has a value which is a reference to a list where each element is a KBaseSearchEngine.SortingRule
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	total has a value which is an int
-	total_in_page has a value which is an int
 	search_time has a value which is an int
+	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -422,8 +422,8 @@ SearchObjectsOutput is a reference to a hash where the following keys are define
 	sorting_rules has a value which is a reference to a list where each element is a KBaseSearchEngine.SortingRule
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	total has a value which is an int
-	total_in_page has a value which is an int
 	search_time has a value which is an int
+	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -580,6 +580,7 @@ MatchValue is a reference to a hash where the following keys are defined:
 GetObjectsOutput is a reference to a hash where the following keys are defined:
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	search_time has a value which is an int
+	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -685,6 +686,7 @@ MatchValue is a reference to a hash where the following keys are defined:
 GetObjectsOutput is a reference to a hash where the following keys are defined:
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	search_time has a value which is an int
+	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -1808,8 +1810,9 @@ Output results for 'search_objects' method.
 'pagination' and 'sorting_rules' fields show actual input for
     pagination and sorting.
 total - total number of found objects.
-total_in_page - total number of objects looked at to get one page of results. 
 search_time - common time in milliseconds spent.
+list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
+    not been updated in search
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
@@ -1834,8 +1837,8 @@ pagination has a value which is a KBaseSearchEngine.Pagination
 sorting_rules has a value which is a reference to a list where each element is a KBaseSearchEngine.SortingRule
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 total has a value which is an int
-total_in_page has a value which is an int
 search_time has a value which is an int
+removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -1851,8 +1854,8 @@ pagination has a value which is a KBaseSearchEngine.Pagination
 sorting_rules has a value which is a reference to a list where each element is a KBaseSearchEngine.SortingRule
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 total has a value which is an int
-total_in_page has a value which is an int
 search_time has a value which is an int
+removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -1915,7 +1918,8 @@ match_filter has a value which is a KBaseSearchEngine.MatchFilter
 =item Description
 
 Output results of get_objects method.
-
+list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
+    not been updated in search
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
@@ -1938,6 +1942,7 @@ mapping<obj_ref, object_info> objects_info - information about each object in th
 a reference to a hash where the following keys are defined:
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 search_time has a value which is an int
+removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -1951,6 +1956,7 @@ objects_info has a value which is a reference to a hash where the key is a KBase
 a reference to a hash where the following keys are defined:
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 search_time has a value which is an int
+removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
--- a/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
+++ b/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
@@ -1807,8 +1807,11 @@ Output results for 'search_objects' method.
     pagination and sorting.
 total - total number of found objects.
 search_time - common time in milliseconds spent.
-list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
-   objects will match input from pagination.
+list<ObjectData> objects - Results of search. In rare cases, due to time delay in updating the search
+  index from  changes in the source data, objects returned from the search
+  index may not be accessible at the data source. The API will remove
+  these objects from the returned search results which means the returned
+  results may have fewer objects than a full page.
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
@@ -1912,7 +1915,10 @@ match_filter has a value which is a KBaseSearchEngine.MatchFilter
 =item Description
 
 Output results of get_objects method.
-list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed.
+list<ObjectData> objects - Results of search. In rare cases, due to time delay in updating the search
+  index from  changes in the source data, objects returned from the search
+  index may not be accessible at the data source. The API will remove
+  these objects from the returned search results.
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
--- a/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
+++ b/lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
@@ -314,7 +314,6 @@ SearchObjectsOutput is a reference to a hash where the following keys are define
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	total has a value which is an int
 	search_time has a value which is an int
-	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -423,7 +422,6 @@ SearchObjectsOutput is a reference to a hash where the following keys are define
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	total has a value which is an int
 	search_time has a value which is an int
-	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -580,7 +578,6 @@ MatchValue is a reference to a hash where the following keys are defined:
 GetObjectsOutput is a reference to a hash where the following keys are defined:
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	search_time has a value which is an int
-	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -686,7 +683,6 @@ MatchValue is a reference to a hash where the following keys are defined:
 GetObjectsOutput is a reference to a hash where the following keys are defined:
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	search_time has a value which is an int
-	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 	access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 	access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 	objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -1811,8 +1807,8 @@ Output results for 'search_objects' method.
     pagination and sorting.
 total - total number of found objects.
 search_time - common time in milliseconds spent.
-list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-    not been updated in search
+list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
+   objects will match input from pagination.
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
@@ -1838,7 +1834,6 @@ sorting_rules has a value which is a reference to a list where each element is a
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 total has a value which is an int
 search_time has a value which is an int
-removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -1855,7 +1850,6 @@ sorting_rules has a value which is a reference to a list where each element is a
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 total has a value which is an int
 search_time has a value which is an int
-removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a KBaseSearchEngine.object_info
@@ -1918,8 +1912,7 @@ match_filter has a value which is a KBaseSearchEngine.MatchFilter
 =item Description
 
 Output results of get_objects method.
-list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
-    not been updated in search
+list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed.
 mapping<access_group_id, narrative_info> access_group_narrative_info - information about
    the workspaces in which the objects in the results reside. This data only applies to
    workspace objects.
@@ -1942,7 +1935,6 @@ mapping<obj_ref, object_info> objects_info - information about each object in th
 a reference to a hash where the following keys are defined:
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 search_time has a value which is an int
-removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
@@ -1956,7 +1948,6 @@ objects_info has a value which is a reference to a hash where the key is a KBase
 a reference to a hash where the following keys are defined:
 objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 search_time has a value which is an int
-removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
 access_group_narrative_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.narrative_info
 access_groups_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.access_group_id and the value is a KBaseSearchEngine.access_group_info
 objects_info has a value which is a reference to a hash where the key is a KBaseSearchEngine.obj_ref and the value is a Workspace.object_info
--- a/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
+++ b/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
@@ -850,9 +850,8 @@ public class WorkspaceEventHandler implements EventHandler, WorkspaceInfoProvide
         // check deletion state
         {
             final StatusEvent updatedEvent = updateEventForDeletion(ev);
-            // if event was updated (event type gets changed to
-            // StatusEventType.DELETE_ALL_VERSIONS
-            if (!updatedEvent.equals(ev) || updatedEvent.getEventType().equals(StatusEventType.DELETE_ALL_VERSIONS)) {
+            // event only checked for deletion. 
+            if (updatedEvent.getEventType().equals(StatusEventType.DELETE_ALL_VERSIONS)) {
                 return updatedEvent;
             }
         }
--- a/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
+++ b/lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
@@ -844,13 +844,20 @@ public class WorkspaceEventHandler implements EventHandler, WorkspaceInfoProvide
             return ev;
         }
 
+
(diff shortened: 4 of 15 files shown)


MrCreosote:
My first major question is why return nulls rather than just removing the data from the returned data structure?

tiramisu24:
So I spoke with Erik and he really really wants the nulls since many parts of the UI will break if they get less objects than expected.

MrCreosote:
That makes no sense. How can the UI expect a certain number of objects to be returned in a search? There may be any number of matches up to the maximum count allowed by the service.

tiramisu24:
From what I understand, the main thing is how results are paged. So there is a total search results number that is used to calculated number of pages. Row size and number of results are determined based on window height. If results are less than whats expected, ui would need to add logic to either 
 1. fetch more results, which could break pagination as number of actual pages could be less than advertised. 
2. back fill nulls so that search results renders normally, with same row heights

tiramisu24:
Basically it seems like the path of least resistance is to return as nulls for a problem that really shouldnt happen that often,

MrCreosote:
Ok, so the UI knows, *after* a given search, how many total results there should be, and also how many results should exist in the current returned data structure.

The only difference is if the returned data structure has nulls in it or fewer elements than expected. It seems to me that the most sensible approach is to just return the data structure with fewer elements. The UI can the do whatever it wants with the last X missing elements - just treat them like nulls, or leave blank rows, or whatever. Regardless, we should not allow one of potentially many consumers of the API to dictate insane API design like using nulls to represent missing data where there's no relationship such that returned data elements can matched to data elements in the input query.

A simple workaround for the UI is:
```
if returned_data.size() < expected_size:
   returned_data.add_all([null] * (expected_size - returned_data.size()))
```
This is a simple, backwards compatible fix and allows the UI to have nulls represent missing data as desired.

MrCreosote:
Also, we should document this edge case in the API documentation.

eapearson:
It is almost a half-dozen or six. I would have preferred if inaccessible result objects are set to null, and the associated object info and workspace info (if relevant) simply omitted, leaving no trace of the inaccessible result object other than the null value in its result list position. The front end would be free to remove them, skip them in the display, or display them differently.

As it is, I will simply create the null entries in the ui code to ensure the result list is of the expected size.  The current code already does this, based on the null value found in the workspace or object info already returned with the search result (set to null if the workspace or object are inaccessible.) A only of this approach in terms of data privacy is that the GUIDs for the inaccessible objects are still revealed -- I think their presence should be completely obscured or removed from the search result. And in terms of the search results representation, the missing items will not be in their correct position, so the ui will most likely insert them at the tail of the result list. This is not terrible important for any search interface we have now, but notable.

Any evaluation should be in the context of how the api is to be used. Where paging is used, as in the current rev of public search, it is important that the search result list length be stable. Paging calculation uses the total search space (returned with each search result, so it may change between pages) and the page size (which is requested, and known only when the results are returned, but if the search space size is greater than the page size, it is expected that the search returns the # of requested objects).

In others, e.g. infinite scrolling, the result list size is less important -- whatever is returned is simply added to a ui-side buffer. It is important to know when the end of the search space is reached, though, so when requesting the final page of data, the numbers need to add up. Thus, knowing that results were dropped is indeed important, so it is good that the dropped objects are included in a new structure.

One tiny quibble with the implementation -- it uses camel case for the field name, existing ones are snake_case. This just makes typing or checking the api more troublesome.

tiramisu24:
@MrCreosote can you review this?

MrCreosote:
> in terms of data privacy is that the GUIDs for the inaccessible objects are still revealed -- I think their presence should be completely obscured or removed from the search result.

I'm in favor of this - that was my original position.

MrCreosote:
> Where paging is used, as in the current rev of public search, it is important that the search result list length be stable.

IMO it's more important that unexpected and unintuitive `null`s don't show up in the results list. That'd mean everyone that uses the API has to special case the `nulls` when usually they just won't care.

I can think of a couple options here:
1) Leave things as is except don't return the GUIDs. Document that in rare cases (once we fix the deletion bug) the page size returned by the API may be less than expected. The client can determine whether there are more pages to be displayed via:
```
more_pages = page_count * page_size < total_records
```

2) Add a `removed_results` (or something) field that `= expected_page_size - page_size`. That's redundant information, but makes things explicit and provides a reminder that `page_size` may fluctuate.

MrCreosote:
@tiramisu24 could you please add/change the title to something descriptive of the purpose of the PR?

MrCreosote:
I suppose we could also make things stupid simple by adding a `more_pages` boolean to the result. Redundant, but...

eapearson:
Regarding nulls in the results, I think a model is batch requests to the workspace when ignoreErrors is present. When one expects that model, then null-handling is natural.

But if we skip them, which is okay, I think we can drop the inaccessible items completely from the workspace and object info and don't need to provide the skipped object guide. The actual return objects, the total available count, and the pagination start and requested page size, all returned with the results, provide enough information.

In either case, providing null for an inaccessible item, or skipping them, the searchapi user will always need to be aware that a result object may be inaccessible and will need to adjust their view model and/or view accordingly.

MrCreosote:
>  Regarding nulls in the results, I think a model is batch requests  to the workspace [e.g. in get_objects2 and get_object_info3 - gavin] when ignoreErrors is present. When one expects that model, then null-handling is natural.

Sure, because there's a 1:1 correspondence between items in the input and items in the output, and the correspondence is maintained by the ordering of the input and output lists.

The reason there are nulls is because ignoreErrors was added after the the api was in widespread use and the nulls were necessary to maintain backwards compatibility. You can't compress the list because that destroys the correspondence.

If I were to design it from scratch I'd probably consider an output mapping instead of a list to avoid the nulls, although the user still has to special case missing output. Another option would be a list item that provides an error code or the data.

Regarding search, there is no 1:1 correspondence so the nulls don't make sense. It'd be like if list_objects returned nulls in the list.

> I think we can drop the inaccessible items completely from the workspace and object info and don't need to provide the skipped object guide.

Again, no arguments here, we should do it.

> In either case, providing null for an inaccessible item, or skipping them, the searchapi user will always need to be aware that a result object may be inaccessible and will need to adjust their view model and/or view accordingly.

Yes, the behavior needs to be clearly documented.

tiramisu24:
@MrCreosote @eapearson How about when seeing inaccessible objects show up actually becomes a small issue, we drop can drop the guide? Right now having the guide will be more useful as every page seems to have some inaccessible object.

MrCreosote:
That kind of stuff should be logged rather than returned in the results. It's not something to which the user should have access.

MrCreosote:
@tiramisu24 could you please add/change the title to something descriptive of the purpose of the PR?

KBaseSearchEngine.spec
```diff
@@ -261,6 +261,8 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
+      list<GUID> removedGuids - list of result GUIDs that are inaccessible or deleted in the workspace but have
```
MrCreosote: I thought we agreed that this would go in the additional properties and we wouldn't officially document it.

KBaseSearchEngine.spec
```diff
@@ -261,6 +261,8 @@ module KBaseSearchEngine {
...
+          not been updated in search. Returned when setting env SHOWREMOVEDGUIDS to anything. Returned in additional properties
```
MrCreosote: Also, we should clearly document the behavior of the search methods and that they may not always return $limit objects, even when there are more than $limit objects to go.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -81,12 +86,17 @@ public SearchTypesOutput searchTypes(final SearchTypesInput params, final String
     public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final String user)
             throws Exception {
         SearchObjectsOutput searchObjsOutput = searchInterface.searchObjects(params, user);
+
+        if(searchObjsOutput.getAdditionalProperties().get(removedGuids) == null && System.getProperty(removedGuidsEnv) != null){
```
MrCreosote: line wrap

MrCreosote: I would require an exact value rather than just non-null for the environment variable to avoid unexpected use. `true` would work. Also namespace the env var a bit like `KBASE_SEARCH_SHOW_REMOVED_GUIDS`.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -41,9 +43,12 @@
  */
 public class NarrativeInfoDecorator implements SearchInterface {
 
+    private final Logger logger = LoggerFactory.getLogger(NarrativeInfoDecorator.class);
```
MrCreosote: Don't make a logger instance. There's some weird effects that I haven't totally figured out, or understand, regarding making a reference to a logger and OOM errors.

Instead do:

```
private Logger getLogger() {
    return LoggerFactory.getLogger(...);
}
```
and use that.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: whitespace around if. Actually, there's lots of inconsistent whitespace - do a pass through the code & fixup

MrCreosote (changes requested):

KBaseSearchEngine.spec
tiramisu24: should document this somewhere though

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: Isn't LoggerFactory.getLogger(...) making an instance?

MrCreosote: Yes sorry, I should've said don't memoize a logger instance in this class, which creates a GC link to the logger instance.

tiramisu24:
@MrCreosote can you review

KBaseSearchEngine.spec
```diff
@@ -261,6 +261,8 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
+      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
```
MrCreosote: I'm having a hard time understanding the second sentence. Also, the explanation should include the eventually consistent angle - this sounds like there's something really stupid about the design.

KBaseSearchEngine.spec
```diff
@@ -305,7 +307,7 @@ module KBaseSearchEngine {
 
     /*
       Output results of get_objects method.
-      
+      list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed.
```
MrCreosote: Same as above.

lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
```diff
@@ -313,8 +313,8 @@ SearchObjectsOutput is a reference to a hash where the following keys are define
 	sorting_rules has a value which is a reference to a list where each element is a KBaseSearchEngine.SortingRule
 	objects has a value which is a reference to a list where each element is a KBaseSearchEngine.ObjectData
 	total has a value which is an int
-	total_in_page has a value which is an int
 	search_time has a value which is an int
+	removedGuids has a value which is a reference to a list where each element is a KBaseSearchEngine.GUID
```
MrCreosote: Needs a recompilation.

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -858,8 +852,7 @@ public StatusEvent updateObjectEvent(final StatusEvent ev)
             final StatusEvent updatedEvent = updateEventForDeletion(ev);
             // if event was updated (event type gets changed to
             // StatusEventType.DELETE_ALL_VERSIONS
-
-            if (!updatedEvent.equals(ev)) {
+            if (!updatedEvent.equals(ev) || updatedEvent.getEventType().name().equals("DELETE_ALL_VERSIONS")) {
```
MrCreosote: Can you explain the logic behind this change?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -81,12 +85,18 @@ public SearchTypesOutput searchTypes(final SearchTypesInput params, final String
     public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final String user)
             throws Exception {
         SearchObjectsOutput searchObjsOutput = searchInterface.searchObjects(params, user);
+        if (searchObjsOutput.getAdditionalProperties().get(removedGuids) == null &&
+                System.getProperty(removedGuidsEnv) != null &&
```
MrCreosote: You can just do `'true'.equals(System.getProperty(removedGuidsEnv)` to avoid the null check

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
     public GetObjectsOutput getObjects(final GetObjectsInput params, final String user)
             throws Exception {
         GetObjectsOutput getObjsOutput = searchInterface.getObjects(params, user);
+        if (getObjsOutput.getAdditionalProperties().get(removedGuids) == null &&
+                System.getProperty(removedGuidsEnv) != null &&
```
MrCreosote: as above

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                        (List<String>) getObjsOutput.getAdditionalProperties().get(removedGuids)));
             }
         }
         return getObjsOutput;
     }
 
+    /**
+     * Adds narrative information for non deleted and valid narrative workspaces. Removes results otherwise and
+     * add the removed result's guid to removedGuids list, if list is not null.
+     * @param objects
```
MrCreosote: Either remove the params and return or fill them out

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                                    .withE1(narrInfo.getNarrativeName())
+                                    .withE2(narrInfo.getNarrativeId())
+                                    .withE3(narrInfo.getTimeLastSaved())
+                                    .withE4(narrInfo.getWsOwnerUsername());
+                    userNames.add(tempNarrInfo.getE4());
+                    retVal.put(workspaceId, tempNarrInfo);
+                }
+                else {
+                    iter.remove();
+                    retVal.put(workspaceId, null);
```
MrCreosote: Why put a null here?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                    if(removedGuids != null) {
```
MrCreosote: space

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                        removedGuids.add(objData.getGuid());
+                    }
+                }
+            //if workspace is not a narrative, remove results from search
+            } else{
```
MrCreosote: I'm confused here. This looks like it's removing any data that does not have a storage code == 'WS'. That seems wrong.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -153,9 +188,17 @@ public GetObjectsOutput getObjects(final GetObjectsInput params, final String us
             }
         }
         catch (IOException | Auth2Exception e) {
-            LoggerFactory.getLogger(getClass()).error("ERROR: Failed retrieving workspace owner realname(s): " +
+            getLogger().error("ERROR: Failed retrieving workspace owner realname(s): " +
                             "setting to null: {}", e.getMessage());
         }
+
+        if(removedGuids != null && removedGuids.size() > 0){
```
MrCreosote: space

MrCreosote: I think we always want to log removed guids, right? So removedGuids should never be null.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +106,76 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                    iter.remove();
+                    retVal.put(workspaceId, null);
+                    if(removedGuids != null) {
+                        removedGuids.add(objData.getGuid());
+                    }
+                }
+            //if workspace is not a narrative, remove results from search
+            } else{
+                iter.remove();
+                if(removedGuids != null) {
```
MrCreosote: Removed guids shouldn't be null so we can log them.

MrCreosote (changes requested):

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: Also don't use the string name to ID deleted events, use the enum class

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: forgot to delete this

KBaseSearchEngine.spec
tiramisu24: How about: 
list<ObjectData> objects - Result of search. Object in transient inaccessible states, such as undergoing deletion or permission change may be included in search total. These inaccessible objects are removed after page is filled.

MrCreosote: I think that'd still be pretty confusing to someone reading the spec for the first time. Maybe something like 

```
Results of search. In rare cases, due to time delay in updating the search
index from  changes in the source data, objects returned from the search
index may not be accessible at the data source. The API will remove
these objects from the returned search results which means the returned
results may have fewer objects than a full page.
```

What do you think?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -81,12 +85,19 @@ public SearchTypesOutput searchTypes(final SearchTypesInput params, final String
     public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final String user)
             throws Exception {
         SearchObjectsOutput searchObjsOutput = searchInterface.searchObjects(params, user);
+        if (searchObjsOutput.getAdditionalProperties().get(removedGuids) == null &&
+                "true".equals(System.getProperty(removedGuidsEnv))){
```
MrCreosote: This line is not fully tested per codecov

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +107,69 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
     public GetObjectsOutput getObjects(final GetObjectsInput params, final String user)
             throws Exception {
         GetObjectsOutput getObjsOutput = searchInterface.getObjects(params, user);
+        if (getObjsOutput.getAdditionalProperties().get(removedGuids) == null &&
+                "true".equals(System.getProperty(removedGuidsEnv))){
```
MrCreosote: This line is not fully tested per codecov

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: this is using system properties - didn't we want to use environment variables rather than system properties?

lib/KBaseSearchEngine/KBaseSearchEngineClient.pm
```diff
@@ -1808,8 +1806,9 @@ Output results for 'search_objects' method.
 'pagination' and 'sorting_rules' fields show actual input for
     pagination and sorting.
 total - total number of found objects.
-total_in_page - total number of objects looked at to get one page of results. 
 search_time - common time in milliseconds spent.
+list<ObjectData> objects - Result of search. Deleted or inaccessible objects are removed. List, before removal of
```
MrCreosote: Looks like this still needs a recompile

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +107,69 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                        getObjsOutput.getAccessGroupNarrativeInfo(),
+                        inputArr));
             }
         }
         return getObjsOutput;
     }
 
+    /**
+     * Adds narrative information for non deleted and valid narrative workspaces. Removes results otherwise and
+     * add the removed result's guid to removedGuids list, if list is not null.
```
MrCreosote: I don't think the list can be null can it?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +107,69 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                    retVal.put(workspaceId, tempNarrInfo);
+                }
+                else {
+                    iter.remove();
+                    removedGuids.add(objData.getGuid());
+
+                }
+            //if workspace is not a narrative, remove results from search
+            } else{
+                iter.remove();
```
MrCreosote: This appears to remove any object where the storage code is not WS, which is not behavior we want - I don't see how it has anything to do with whether a workspace is a narrative or not.

Also, it seems that if we don't want to show non-narrative workspaces in search results (which I think we do, for refdata if nothing else) we should just not index them.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +107,69 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+                            new Tuple5<String, Long, Long, String, String>()
+                                    .withE1(narrInfo.getNarrativeName())
+                                    .withE2(narrInfo.getNarrativeId())
+                                    .withE3(narrInfo.getTimeLastSaved())
+                                    .withE4(narrInfo.getWsOwnerUsername());
+                    userNames.add(tempNarrInfo.getE4());
+                    retVal.put(workspaceId, tempNarrInfo);
+                }
+                else {
+                    iter.remove();
```
MrCreosote, line 167: What about refdata?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -153,9 +182,17 @@ public GetObjectsOutput getObjects(final GetObjectsInput params, final String us
             }
         }
         catch (IOException | Auth2Exception e) {
-            LoggerFactory.getLogger(getClass()).error("ERROR: Failed retrieving workspace owner realname(s): " +
+            getLogger().error("ERROR: Failed retrieving workspace owner realname(s): " +
```
MrCreosote, line 186: I didn't notice this before, but this seems like weird behavior to me - it's returning partial results on a failure, which is really a pain for users of the API. Maybe not in this PR, but in a future PR, it should either retry or throw an error IMO.

MrCreosote (changes requested):

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: Ref data isn't checked. This is for the case of narrative search where narrativeinfo comes back as null, the results are removed.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: I think I missed the conversation about ref data. I can remove this.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: ops didn't update this comment from when we changed to not do nulls.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: Its probably b/c additional properties always doesnt start with removed guids?

tiramisu24: From what I understand, they both work for this case. Since javadoc recommended I use system properties, I went with that. Also there were other parts of the code that used the system properties, I figured we wanted to keep it consistent.

MrCreosote: I don't understand - whatever's in additional properties is determined by your tests.

MrCreosote: Can you verify that environment variables show up in the system properties? I didn't think that was the case.

> Since javadoc recommended I use system properties, I went with that.

What javadoc are you talking about?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: Right, and for refdata I presume narrative info will be null, so all refdata will be removed from the results, right?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#EnvironmentVSSystemProperties
I don't think env variables show up in system properties? Also suppose if we were to use environment variables, how would you set/unset the env variable in a test?

tiramisu24: what would be the point of having an additional two cases in which additional properties have removedGuids? Right now they never should start with the removed guids

MrCreosote: If that's the case why is the code checking for a case that can never happen?

MrCreosote: So if we do want to use env vars then we need changes here. I don't have strong feelings but if we don't use env vars then there needs to be more engineering to set the sys props on startup. Env vars would be easier to deal with.

https://github.com/kbase/KBaseSearchEngine/blob/master/test/src/kbasesearchengine/test/common/TestCommon.java#L266

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: Good point. How can I tell if an object is a refdata?

tiramisu24: are sourcetags reliable?

MrCreosote: Actually, under what cases is narrativeInfo null?

MrCreosote: The reason I ask is that this logic was already there based on the diff, and I assume we're not excluding refdata now.

tiramisu24: @Override
    public NarrativeInfo findNarrativeInfo(final Long wsId) {
        try {
            return cache.get(wsId);
        } catch (Exception e) {
            LoggerFactory.getLogger(getClass()).error("ERROR: Failed retrieving narrative info: Returning null:  {}",
                    e.getMessage());
            // unchecked exceptions are wrapped in UncheckedExecutionException
            return null;
        }
    }

tiramisu24: public NarrativeInfo findNarrativeInfo(final Long accessGroupID) {
        Tuple9 <Long, String, String, String, Long, String, String,
                String, Map<String,String>> wsInfo;

        try {
            wsInfo = wsHandler.getWorkspaceInfo(accessGroupID);
        } catch (IOException | JsonClientException e) {
            LoggerFactory.getLogger(getClass()).error("ERROR: Failed retrieving workspace info: Returning null: {}",
                    e.getMessage());
            wsInfo = null;
        }

        if (wsInfo == null) return null;

        final long timeMilli = WorkspaceEventHandler.parseDateToEpochMillis(wsInfo.getE4());

        final NarrativeInfo tempNarrInfo =
                new NarrativeInfo(null, null, timeMilli, wsInfo.getE3());

        final Map<String, String> wsInfoMeta = wsInfo.getE9();

        if (wsInfoMeta.containsKey("narrative") &&
                wsInfoMeta.containsKey("narrative_nice_name")) {
            tempNarrInfo.withNarrativeName(wsInfoMeta.get("narrative_nice_name"))
                    .withNarrativeId(Long.parseLong(wsInfoMeta.get("narrative")));
        }
        return tempNarrInfo;
    }

tiramisu24: ugh can't get the formatting right

tiramisu24: looks like if only returns nulls if workspace throughs error. For a refdata workspace, would the rest of code error?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: That link for changing env vars seems super hacky. What if we do something where, on startup, we have a check where if the env variable is set, then setup the System properties? And then the tests can do whatever with the system properties.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: This also seems odd, since it means the results returned from the search API may be different over time.

But in any case, non-narrative results, based on this code, are not removed.

> For a refdata workspace, would the rest of code error?

Do you see any reason it would?

tiramisu24: what do you mean different over time? 

idk. would fields from the wsInfo such as E9 (metadata) ever be null?

MrCreosote: Wait a sec, this makes no sense. If there's a network blip and the workspace can't be contacted then no data will be returned.

tiramisu24: You mean wsInfo would be empty? wsHandler wouldn't through an error?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: > That link for changing env vars seems super hacky.

It is, but it's just for tests, so who cares.

> What if we do something where, on startup, we have a check where if the env variable is set, then setup the System properties?

Why not just check the environment instead of the sys props?

tiramisu24: omg I think that hack is doing is putting the env into properties. A search on the project shows that the only thing it adds is KB_DEPLOYMENT_CONFIG. To get local KBaseSearch up, I had to add System.setProperty("KB_DEPLOYMENT_CONFIG", dir);

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: if you look at the code above, if *any* error occurs contacting the ws, narrinfo is set to null, which means the data is removed from the results returned by the search API.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: The hack is used for setting up the environment in tests so the API can be started.

I'm not sure what you mean about local KBaseSearch. To run the the server, you have to set the deployment config env var appropriately.

MrCreosote: For example https://github.com/kbase/KBaseSearchEngine/blob/master/Dockerfile.release#L17

tiramisu24: It's okay. Since I don't really have a preference, i can just do it this way

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: isn't that what we want? If there is a network blip and we won't get any data, the we can't show the data on the front end anyway.

tiramisu24: I guess we can do retries. but if we are going to do that, there are so many other areas that could use this...

MrCreosote: > isn't that what we want? If there is a network blip and we won't get any data, the we can't show the data on the front end anyway.

No, the difference here is that the user will think there's no results for their search rather than there was an error with their search. If there's a blip we should retry or throw an error, not tell the user there aren't any results.

tiramisu24: what would no results look like?

MrCreosote: Depends on the UI. For the API an empty list of results.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: doing this in another pr

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: I think this fix is going in another PR

tiramisu24: Should be in this pr. Right now for this pr we are assuming that a null result from a workspace means it has been deleted. The next Pr should be handling the different errors (deletion vs non deletion). And if there needs a different check for handling deleted workspaces, it should be changed later.

MrCreosote: > The next Pr should be handling the different errors (deletion vs non deletion)

that's the fix I meant. The original issue isn't an issue if we assume that the narrative info provider will be fixed such that it only returns null if the narrative is deleted.

MrCreosote: Although we should probably document - very clearly - what's going on here and when the narrative info provider is supposed to return null and when it isn't.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +105,64 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
-                        getObjsOutput.getAccessGroupNarrativeInfo()));
+                        getObjsOutput.getAccessGroupNarrativeInfo(),
+                        inputArr));
             }
         }
         return getObjsOutput;
     }
 
+    /**
+     * Adds narrative information for non deleted and valid narrative workspaces. Removes results otherwise and
```
MrCreosote: Just to be sure - my understanding is that this code makes no distinction between narrative and non-narrative workspaces, right?

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
```diff
@@ -858,8 +852,7 @@ public StatusEvent updateObjectEvent(final StatusEvent ev)
             final StatusEvent updatedEvent = updateEventForDeletion(ev);
             // if event was updated (event type gets changed to
             // StatusEventType.DELETE_ALL_VERSIONS
-
-            if (!updatedEvent.equals(ev)) {
+            if (!updatedEvent.equals(ev) || updatedEvent.getEventType().equals(StatusEventType.DELETE_ALL_VERSIONS)) {
```
MrCreosote: I'm sorry, but I've forgotten the driver for this change again. What was the issue this solves?

Also, I think there's another bug here - if a deletion event comes in but the object has since been undeleted, the deletion event should be updated to an undelete, right?

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -121,12 +125,12 @@ public void searchObjectsDecorateWithNullInfo() throws Exception {
         final SearchObjectsInput dummyInput = new SearchObjectsInput()
                 .withPostProcessing(new PostProcessing().withAddNarrativeInfo(1L));
 
-        final List<ObjectData> objectdata = Arrays.asList(
+        final List<ObjectData> objectdata = new ArrayList<>(Arrays.asList(
```
MrCreosote, line 128: Why is the object data wrapped in another list? I don't see any code changes that would necessitate that.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +105,64 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
-                                .withE4(narrInfo.getWsOwnerUsername());
-                userNames.add(tempNarrInfo.getE4());
-                retVal.put(workspaceId, tempNarrInfo);
-            }
-            else {
-                retVal.put(workspaceId, null);
+                final long workspaceId = guid.getAccessGroupId();
+                final NarrativeInfo narrInfo = narrInfoProvider.findNarrativeInfo(workspaceId);
+
+                if (narrInfo != null) {
```
MrCreosote, line 156: Comment here noting that narrInfo is only null if the workspace is deleted. If that's not documented in the provider class should add docs in the PR that fixes the bug there.

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +376,124 @@ public void listTypes() throws Exception {
...
+            expectedObjs.add(new ObjectData().withGuid("WS:4/1/7").withCreator("user"));
+            expectedObjs.add(new ObjectData().withGuid("WS:4/2/1").withCreator("user"));
+
+            env.put("KBASE_SEARCH_SHOW_REMOVED_GUIDS", "true");
+            final SearchObjectsOutput res = nid.searchObjects(dummyInput, "user");
+
+            compareSearchObjectOutputRes(res.getObjects(), expectedObjs);
+            compare( res.getAccessGroupNarrativeInfo().get(4L), narrInfoTuple("test", 1L, 1L, "user", null));
+
+            assertThat("removed guids shown", res.getAdditionalProperties().get("removedGuids"), is(Arrays.asList("WS:3/5/6")));
```
MrCreosote: line wrap

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +376,124 @@ public void listTypes() throws Exception {
...
+            expectedObjs.add(new ObjectData().withGuid("WS:4/1/7").withCreator("user"));
+            expectedObjs.add(new ObjectData().withGuid("WS:4/2/1").withCreator("user"));
+
+            env.put("KBASE_SEARCH_SHOW_REMOVED_GUIDS", "true");
+            final GetObjectsOutput res = nid.getObjects(dummyInput, "user");
+
+            compareSearchObjectOutputRes(res.getObjects(), expectedObjs);
+            compare( res.getAccessGroupNarrativeInfo().get(4L), narrInfoTuple("test", 1L, 1L, "user", null));
+
+            assertThat("removed guids shown", res.getAdditionalProperties().get("removedGuids"), is(Arrays.asList("WS:3/5/6")));
```
MrCreosote: line wrap

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -381,6 +503,18 @@ public static void compare(
         }
     }
 
+    public static void compareSearchObjectOutputRes(List<ObjectData> expected, List<ObjectData> res ){
+        assertThat("incorrect number of results", expected.size(), is(res.size()));
+        for(int i =0; i<Math.min(expected.size(), res.size()); i++){
```
MrCreosote: spaces

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -381,6 +503,18 @@ public static void compare(
...
+            if(res.get(i) == null){
+                assertThat("incorrect object", expected.get(i) == null, is(true));
+            }else{
```
MrCreosote: spaces

MrCreosote (changes requested):

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: But the original issue is that we still want deleted objects to be removed from in the results?

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
tiramisu24: b/c originally I was suppose to set results to null, or to keep calling searchAPI until I fielded enough results. At one point Arrays.asList didn't like me mutating the content so I changed all the tests

MrCreosote: Ok, can you put them back the way they were then? Or does that still cause problems?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: The original issue from the top comment in this thread is

> What about refdata?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: Yes. This was my misunderstanding

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: refdata would still be included? it should still have workspace info correct?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: hmm weird I thought I deleted this already

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
tiramisu24: what would an undelete event be?

MrCreosote: https://github.com/kbase/KBaseSearchEngine/blob/83b7f797f3ed221c725e7b9b384a2c2eeee83632/lib/src/kbasesearchengine/events/StatusEventType.java#L16

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: Ok, can you update the comment please?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: I don't really understand what you're asking here.

My understanding is that previous to this PR, refdata would not be excluded from the returned results because the narrativeInfoProvider doesn't actually provide just narrative info. It provides info about the workspace, and also provides additional info about the narrative if the workspace is a narrative. So narrInfo is only null if the workspace isn't accessible, not if the workspace is not a narrative, which is unintuitive, which is why I thought refdata would be removed from the result. 

Since my intuition was wrong, I don't think the original concern about refdata in the original comment was an issue, and nothing in this PR changes that.

What we should do is 

> document - very clearly - what's going on here and when the narrative info provider is supposed to return null and when it isn't.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: You did, this is a 9 day old comment.

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
tiramisu24: do I have to?

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: okay will add comments explaining this

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
MrCreosote: One of the cardinal sins of SW development is adding unnecessary complexity so I'm going with yeah :P

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
tiramisu24: do you want me to make the undelete changes to this pr though? also the line could just be:

if (updatedEvent.getEventType().equals(StatusEventType.DELETE_ALL_VERSIONS)

so that if the event is deleted, it doesnt do all the shenanigans to add to the event (which also makes workspace calls)

MrCreosote: No, just add the bug to the list.

I think updateEventForDeletion will *only* update to a delete event, right? If so, it might be more clear only return an event if it's update and return `null` (or `Optional.absent()`) otherwise. Then the check is `if (!updatedEvent == null | isDeletedEvent()) {` which is more readable.

I don't really have a super strong preference though.

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +105,64 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+        final Iterator<ObjectData> iter = objects.iterator();
+        while (iter.hasNext()) {
+            final ObjectData objData = iter.next();
             final GUID guid = new GUID(objData.getGuid());
             if (WorkspaceEventHandler.STORAGE_CODE.equals(guid.getStorageCode())) {
-                wsIdsSet.add((long) guid.getAccessGroupId());
-            }
-        }
-        final Set<String> userNames = new HashSet<>();
-        for (final long workspaceId: wsIdsSet) {
```
MrCreosote, line 128: Why is the 2 stage loop removed? That'll result in up to  9999 more calls to the narrative info provider than necessary.

MrCreosote (changes requested):

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: narrative info provider is a cache with size 50000 * 1000 (why is it so large??), so it should be fine. Otherwise, need to keep track of deleted ws ids, and then go through the again to remove all the objects in deleted ws.

MrCreosote: No idea why it's so large. 50k should be more than enough I'd think.

That being said, that's an implementation detail, and this code has no idea how the provider is implemented, and the provider could be swapped out at any time, so relying on implementation details of narrative info provider is not something you want to do.

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
tiramisu24: actually it does still cause problems as Arrays.asList won't let me remove items. https://stackoverflow.com/questions/7885573/remove-on-list-created-by-arrays-aslist-throws-unsupportedoperationexception

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
tiramisu24: what do do if it is a non deletion exception then? Right now that also returns null, but idk what we would want to change that too

MrCreosote: If it's not a deletion exception then the exception should be thrown (or possibly wrapped) and thrown, as the deleted exception is the only exception we should expect under normal circumstances, AFAIK

lib/src/kbasesearchengine/events/handler/WorkspaceEventHandler.java
MrCreosote: This ? is still unanswered I think:

> I'm sorry, but I've forgotten the driver for this change again. What was the issue this solves?

tiramisu24: ops fixed

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -44,6 +46,8 @@
     private final SearchInterface searchInterface;
     private final NarrativeInfoProvider narrInfoProvider;
     private final AuthInfoProvider authInfoProvider;
+    private final String removedGuids = "removedGuids";
```
MrCreosote: These should both be static with ALLCAPS variable names since they're constants

KBaseSearchEngine.spec
```diff
@@ -261,6 +261,11 @@ module KBaseSearchEngine {
           pagination and sorting.
       total - total number of found objects.
       search_time - common time in milliseconds spent.
+      list<ObjectData> objects - Results of search. In rare cases, due to time delay in updating the search
```
MrCreosote, line 264: Needs release notes

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
MrCreosote: All the other API properties are under_score rather than camelCase, so this one should be as well

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -96,51 +105,72 @@ public SearchObjectsOutput searchObjects(final SearchObjectsInput params, final
...
+
+                if(seenWorkspaces.contains(workspaceId)){
+                    continue;
+                }
+
+                seenWorkspaces.add(workspaceId);
+                final NarrativeInfo narrInfo = narrInfoProvider.findNarrativeInfo(workspaceId);
+
+                //provider sets narrative info null for deleted workspaces
+                if (narrInfo != null) {
```
MrCreosote, line 156: Noting that narrInfo will be null if *any* problem whatsoever occurs getting the workspace info from the workspace. The fix for this is out of scope for this PR.

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +376,128 @@ public void listTypes() throws Exception {
...
+            expectedObjs.add(new ObjectData().withGuid("WS:4/2/1").withCreator("user"));
+
+            env.put("KBASE_SEARCH_SHOW_REMOVED_GUIDS", "true");
+            final GetObjectsOutput res = nid.getObjects(dummyInput, "user");
+
+            compareSearchObjectOutputRes(res.getObjects(), expectedObjs);
+            compare( res.getAccessGroupNarrativeInfo().get(4L),
+                    narrInfoTuple("test", 1L, 1L, "user", null));
+
+            assertThat("removed guids shown", res.getAdditionalProperties().get("removedGuids"), is(Arrays.asList("WS:3/5/6")));
```
MrCreosote: line wrap

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -381,6 +507,18 @@ public static void compare(
         }
     }
 
+    public static void compareSearchObjectOutputRes(List<ObjectData> expected, List<ObjectData> res ){
+        assertThat("incorrect number of results", expected.size(), is(res.size()));
+        for (int i =0; i < Math.min(expected.size(), res.size()); i++){
+            if (res.get(i) == null) {
+                assertNull(expected.get(i));
+            }else{
```
MrCreosote: spaces

test/src/kbasesearchengine/test/main/NarrativeInfoDecoratorTest.java
```diff
@@ -372,6 +376,128 @@ public void listTypes() throws Exception {
...
+
+        final ArrayList<ObjectData> objs = new ArrayList<>();
+        objs.add(obj);
+        objs.add(obj2);
+        objs.add(obj3);
+
+        when(nip.findNarrativeInfo(3L)).thenReturn(null);
+        when(nip.findNarrativeInfo(4L)).thenReturn(new NarrativeInfo("test", 1L, 1L, "user"));
+
+        when(search.searchObjects(Mockito.any(SearchObjectsInput.class), eq("user"))).thenReturn(new SearchObjectsOutput()
```
MrCreosote: line wrap

MrCreosote (changes requested):
Some minor quick fixes, then it's ready to go.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,11 @@
 Search Service MKII release notes
 =================================
+0.2.4
```
MrCreosote, line 3: is 0.2.3 released already? if not just add to the 0.2.3 notes

lib/src/kbasesearchengine/main/NarrativeInfoDecorator.java
```diff
@@ -44,6 +46,8 @@
     private final SearchInterface searchInterface;
     private final NarrativeInfoProvider narrInfoProvider;
     private final AuthInfoProvider authInfoProvider;
+    private final static String REMOVED_GUIDS = "REMOVED_GUIDS";
```
MrCreosote: While the variable name should be caps, the string should follow the same style as the spec, so 'removed_guids'.

MrCreosote (changes requested):
1 last thing & 1 question

RELEASE_NOTES.md
tiramisu24: 0.2.3 already released

MrCreosote (approved):

PR #271, opened 2018-08-30
Fix SCT-1324: search_types doesn't return more than 10 types

eapearson:
The bug is that search_types was only returning at most 10 types. This is due to the ES default behavior for term aggregations. A "size" field added to the term fixes the problem.

The value for this is set in ElasticIndexingStorage.java at 1000, which can be overridden by a constructor which takes this value as a new argument. This value was chosen based advice that one can choose a size limit an order of magnitude greater than the current estimated scale of on the number of unique terms being aggregated over.  We currently have around 20 types, so round that safely up to 100, then choose 1000.

While writing the tests for this, it was discovered that the searchTypes call fails if the indexes are empty, so a fix was put in place for that as well.

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -184,7 +184,9 @@ public class ElasticIndexingStorage implements IndexingStorage {
         this.esHost = esHost;
         this.indexNamePrefix = "";
         this.tempDir = tempDir;
-        this.searchTypesAggregationSize = aggregationSize;
+        if (aggregationSize != null) {
+            this.searchTypesAggregationSize = aggregationSize;
+        }
     }
     
     public HttpHost getEsHost() {
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -174,12 +174,46 @@ public class ElasticIndexingStorage implements IndexingStorage {
      */
     public static final int MAX_OBJECT_TYPES_SIZE = 50;
 
+    /**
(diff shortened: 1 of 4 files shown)


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -165,6 +179,15 @@ public ElasticIndexingStorage(HttpHost esHost, File tempDir) throws IOException
         this.indexNamePrefix = "";
         this.tempDir = tempDir;
     }
+
```
bio-boris, line 195: @eapearson Could you please put a javadoc for this function? Also we should probably have javadocs for other functions as well in this file.

eapearson: Thanks Boris. I added javadocs for the constructors. Can we add docs for the other functions later - you can even assign it to me! - ? I agree they need them -- I had to look far and wide to find examples of Javadoc practices in the codebase! This bug is exposed in the searchapi, so I just want to close the loop on it if everything looks kosher.

bio-boris (approved):
Tests pass,

eapearson:
For posterity - I ran the new tests many times, and the entire test suite several times, using the host (Mac) not kb-sdk (docker). Could never get the entire test suite to go in one pass with kb-sdk test, although I could get all test to pass if I split them up. I think this is a docker resource constraint issue -- my Mac doesn't have too much memory to spare but giving docker more does help. In order to work at all, kb-sdk test required modifying the Dockerfile to use a non-root user. Using docker-compose or similar orchestration tool would obviate that change, since that was to accommodate elasticsearch, and that also might make the tests happier. Still running as non-root is much better anyway.

eapearson:
@bio-boris possible to get this merged?

PR #275, opened 2018-09-12
Upgrade ES from 5.5 to 5.6.11

bio-boris:
Upgrade ES from 5.5 to 5.6.11

### DIFF
--- a/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
+++ b/lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
@@ -726,7 +726,7 @@ public class ElasticIndexingStorage implements IndexingStorage {
         // script = {"inline": "ctx._source.islast = (ctx._source.version == params.lastver)",
         //           "params": {"lastver": lastVersion}}
         Map<String, Object> script = ImmutableMap.of(
-                "inline", "ctx._source.islast = (ctx._source.version == params.lastver);",
+                "source", "ctx._source.islast = (ctx._source.version == params.lastver);",
                 "params", params);
 
         // doc = {"query": {"bool": {"filter": [{"term": {"prefix": prefix}}]}},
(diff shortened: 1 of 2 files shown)


bio-boris:
Note the one error and warnings are already in the current builds

https://travis-ci.org/kbase/KBaseSearchEngine/jobs/427889488#L1656

https://travis-ci.org/kbase/KBaseSearchEngine/jobs/427889488#L1669-L1671

(You might need to click the link again in order to get to the line number, or refresh the page)

MrCreosote:
@tiramisu24 can you review this one please?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -726,7 +726,7 @@ private int updateLastVersionsInData(String indexName, GUID parentGUID,
         // script = {"inline": "ctx._source.islast = (ctx._source.version == params.lastver)",
```
tiramisu24: comments should be changed to source too

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -726,7 +726,7 @@ private int updateLastVersionsInData(String indexName, GUID parentGUID,
...
         //           "params": {"lastver": lastVersion}}
         Map<String, Object> script = ImmutableMap.of(
-                "inline", "ctx._source.islast = (ctx._source.version == params.lastver);",
```
tiramisu24: is inline to source a breaking change?

tiramisu24: actually is this change necessary? I don't see it in breaking changes from 5.5 to 5.6
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/breaking-changes-5.6.html

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
bio-boris: done

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
bio-boris: The change is not necessary, but it makes all of the warnings disappear and thus the logs are not flooded.

PR #276, opened 2018-09-13
SCT-1362

tiramisu24:
-Change check for parent guid to look at object type instead of all of elastic


lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -547,6 +548,29 @@ public void flushIndexing(final ObjectTypeParsingRules rule) throws IOException
         return ImmutableMap.copyOf(ret);
     }
 
+    public boolean lookupParentId(final String type, final GUID guid) throws IOException {
```
bio-boris: Javadoc would be nice here

lib/src/kbasesearchengine/search/IndexingStorage.java
```diff
@@ -58,6 +58,8 @@ public void indexObjects(
     public Map<GUID, Boolean> checkParentGuidsExist(Set<GUID> parentGuids)
             throws IOException;
 
+    public boolean lookupParentId(String type, GUID guid) throws IOException;
```
bio-boris: Javadoc would be nice here

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -424,8 +424,8 @@ private void processEvent(final StatusEvent ev)
             switch (updatedEvent.getEventType()) {
             case NEW_VERSION:
                 GUID pguid = updatedEvent.toGUID();
-                boolean indexed = indexingStorage.checkParentGuidsExist(new LinkedHashSet<>(
-                        Arrays.asList(pguid))).get(pguid);
+                boolean indexed = indexingStorage.lookupParentId(
+                        updatedEvent.getStorageObjectType().get().getType(), updatedEvent.toGUID());
```
bio-boris: can you just pass in pguid, instead of `updatedEvent.toGUID()` ?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
```diff
@@ -547,6 +548,29 @@ public void flushIndexing(final ObjectTypeParsingRules rule) throws IOException
...
+        final String indexName = indexNamePrefix + type + "*";
+        final Set<GUID> guids = new HashSet<>();
+        guids.add(guid);
+        Map<String, Object> doc =
```
bio-boris, line 615: Could you add a comment of the actual query here?
e.g.   like for checkParentGuidsExist    // doc = {"query": {"bool": {"filter": [{"term": {"prefix": prefix}}]}}}

lib/src/kbasesearchengine/main/IndexerWorker.java
tiramisu24: yes, yes I can

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: what should be in there though? This method is actually just look up if parent id exists, given a type and a guid

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
```diff
@@ -1815,4 +1815,26 @@ public void testDeleteObjectWithSubobject() throws Exception {
...
+                SourceData.getBuilder(new UObject(new HashMap<>()), "objname", "creator").build(),
+                Instant.ofEpochMilli(10000),
+                null,
+                new GUID("WS:2000/1/1"),
+                ImmutableMap.of(new GUID("WS:2000/1/1"), new ParsedObject(
+                        "{\"whee\": \"imaprettypony\"}",
+                        ImmutableMap.of("whee", Arrays.asList("imaprettypony")))),
+                false);
+        final boolean res = indexStorage.lookupParentId("someobject", new GUID("WS:2000/1/1"));
+        assertThat("could not find parent id", res, is(true));
```
bio-boris, line 1840: What I think this test is doing:

Step1) Index an object 
Step2) Check to see if parentID of that guid exists. If its true, that means it has been indexed already.

Should we also have a test that checks to see if an event has not been indexed already?

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
bio-boris: Maybe it should be called lookupParentIdExists or hasParentId? I would expect a lookupParentId to retrieve the parentID of the guid. 
The type has to be in a list of predetermined types or can it be an arbitrary type? 
The GUID is the guid of an updatedEvent right, or can it be any arbitraryGuid? The answer to those  might be able to go in the comments

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
tiramisu24: There is already a test in IndexerWorkerTest for that

lib/src/kbasesearchengine/search/ElasticIndexingStorage.java
tiramisu24: hasParentId that makes sense. what is an arbitraryGuid?

bio-boris: like it could be an UpdateEventGuid vs a generic Guid

test/src/kbasesearchengine/test/search/ElasticIndexingStorageTest.java
bio-boris: Ok great

bio-boris (approved):

scanon:
@tiramisu24 I don't think this PR will still address my issue.  You need to check on a specific index not just a type.  This also assumes that the index name is also some variant of the type name.

Imagine the following state...

prod.genomev1 has a record for WS:1/2/3.

A new index is added called ganomev2 and a re-index is kicked off.  We want the code to ignore the fact that WS:1/2/3 exist in prod.genomv1 and instead check explicitly for WS:1/2/3 in prod.genomev2.

How about starting with a test case that creates this scenario and confirm that the check returns the right result.

tiramisu24:
@scanon I think I am misunderstanding something here. I thought that the storageObjectType was the name of the index. That looks like how objects are getting indexed at least. Will add an integration test

PR #277, opened 2018-09-19
SCT-1367

tiramisu24:
-Fix bug with error handling
-Add handling for deleted objects/workspace errors

### DIFF
--- a/lib/src/kbasesearchengine/main/AccessGroupNarrativeInfoProvider.java
+++ b/lib/src/kbasesearchengine/main/AccessGroupNarrativeInfoProvider.java
@@ -42,15 +42,23 @@ public class AccessGroupNarrativeInfoProvider implements NarrativeInfoProvider {
     @Override
     public NarrativeInfo findNarrativeInfo(final Long accessGroupID) {
         Tuple9 <Long, String, String, String, Long, String, String,
-                String, Map<String,String>> wsInfo;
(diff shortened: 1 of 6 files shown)


lib/src/kbasesearchengine/main/AccessGroupNarrativeInfoProvider.java
```diff
@@ -42,15 +42,23 @@ public AccessGroupNarrativeInfoProvider(final WorkspaceEventHandler wsHandler) {
...
         Tuple9 <Long, String, String, String, Long, String, String,
-                String, Map<String,String>> wsInfo;
+                String, Map<String,String>> wsInfo = null;
+
+            try {
+                wsInfo = wsHandler.getWorkspaceInfo(accessGroupID);
+            } catch (IOException e) {
+                if (e.getMessage().toLowerCase().contains("is deleted") ||
+                        e.getMessage().toLowerCase().contains("has been deleted")) {
+                    wsInfo = null;
```
bio-boris: Correct me if I'm wrong, but I think setting wsInfo to null again is redundant, since its set to null when its first defined..

bio-boris (approved):
Error logic handling updates are good. I don't understand what the coverage report means though.

bio-boris (changes requested):
Waiting for tests for "getObjectInfo" before merge

bio-boris (approved):

PR #280, opened 2018-09-27
Update Dockerfile to upgrade search

bio-boris:


bio-boris:
See PR 281 https://github.com/kbase/KBaseSearchEngine/pull/281

PR #283, opened 2018-10-02
SCT-1367 pt2

tiramisu24:
- Continuation of previous PR for SCT-1367
- Change to throw exceptions, instead of catching, logging, and returning null
- Only return null for delete exceptions


tiramisu24:
@bio-boris could you look at this?

PR #284, opened 2018-10-04
New SCT-1362

tiramisu24:


MrCreosote:
it'd be really helpful if the title (and body, if the pr is complex) described the purpose of the pr.

lib/src/kbasesearchengine/events/exceptions/Retrier.java
```diff
@@ -110,6 +112,10 @@ public RetryLogger getLogger() {
                 } else {
                     retries++;
                 }
+            } catch (IOException | IndexingConflictException e) {
```
MrCreosote: The Retrier should definitely not retry non-retriable errors.

lib/src/kbasesearchengine/main/IndexerWorker.java
```diff
@@ -537,23 +525,46 @@ private void indexObject(
...
-                logger.timeStat(guid, 0, parsingTime, indexTime);
+
+                final SearchObjectType globalIdx = rule.getGlobalObjectType();
+
+                //pass into indexingStorage
+                boolean indexed = indexingStorage.hasParentId(globalIdx, guid);
+
+                if (indexed && !overwriteExistingData) {
+                    logger.logInfo("[Indexer]   skipping " + guid +
+                            " creation (already indexed and overwriteExistingData flag is set to false)");
```
MrCreosote, line 536: this message should include the search object type since it could be true for one SOT and not another SOT

lib/src/kbasesearchengine/system/TypeFileStorage.java
```diff
@@ -62,13 +63,16 @@
                 typeToFile.put(searchType, file);
                 searchTypes.put(searchType, new ArrayList<>(types));
                 final CodeAndType cnt = new CodeAndType(types.get(0).getStorageObjectType());
+                final TypeMapping typeMapping = TypeMapping.getBuilder(cnt.storageCode, cnt.storageType)
```
MrCreosote: I don't understand the purpose of this change. Previously there'd be a mapping from a storage type to one or more search types in a `TypeMapping`. Now they're split up into multiple `TypeMapping`s, but it's the same information.

lib/src/kbasesearchengine/system/TypeFileStorage.java
```diff
@@ -159,6 +163,8 @@ public TypeFileStorage(
         Utils.nonNull(fileLister, "fileLister");
         Utils.nonNull(logger, "logger");
         storageTypes = processTypesDir(typesDir, searchSpecParser, fileLister, logger);
+
+        //not sure what the purpose of this is. We could just pair the types together based on storage type
```
MrCreosote: If 'this' means the type mappings, then the purpose of the type mappings is to map specific versions of a storage type against specific versions of the search type. Otherwise there's no way to determine which version of a search type to pick given a storage type. Recall that storage type versions are not equivalent across environments.

lib/src/kbasesearchengine/system/TypeFileStorage.java
```diff
@@ -35,18 +35,19 @@
     // as opposed to file types for mappings
     private static final Set<String> ALLOWED_FILE_TYPES_FOR_TYPES =
             new HashSet<>(Arrays.asList(".json", ".yaml"));
-    
+    //search global type to parsing rules (spec)
```
MrCreosote: As far as I can tell the changes to this file should be reverted.

MrCreosote (changes requested):

lib/src/kbasesearchengine/system/TypeFileStorage.java
tiramisu24: From my understanding this file needs to change. The storagetypes is mapping a kbase storage type (i.e. KBaseGenomes.Genome) to a type mapping (from yaml file). Since we are now doing multiple spec, such as GenomeA.yml, GenomeB.yml, and kbase storage type is still the same, the storageTypes needs to change to an List. Otherwise GenomeB would just override GenomeA

lib/src/kbasesearchengine/system/TypeFileStorage.java
tiramisu24: not the same. There are multiple typemappings for each version of the spec. I thought about adding the search object type into the type mapping but it is redundant b/c we always have a default that has that information. Of course this was based on my understanding of how the specs would work, which may be incorrect...

lib/src/kbasesearchengine/events/exceptions/Retrier.java
tiramisu24: ops. came from intellij auto correct. will revert

lib/src/kbasesearchengine/system/TypeFileStorage.java
MrCreosote: > not the same. There are multiple typemappings for each version of the spec.

Which spec? 

> I thought about adding the search object type into the type mapping but it is redundant b/c we always have a default that has that information.

It is in the type mapping: https://github.com/kbase/KBaseSearchEngine/blob/31227c0e499b129bea0060fad7fbfce2bf4c9aed/lib/src/kbasesearchengine/system/TypeMapping.java#L23-L26

lib/src/kbasesearchengine/system/TypeFileStorage.java
MrCreosote: I don't think that's the case: https://github.com/kbase/KBaseSearchEngine/blob/31227c0e499b129bea0060fad7fbfce2bf4c9aed/lib/src/kbasesearchengine/system/TypeMapping.java#L205-L215

lib/src/kbasesearchengine/system/TypeFileStorage.java
tiramisu24: The storageType is refering to kbase storage type. Not the search storage type. by spec, I meant the yaml files we use to index

lib/src/kbasesearchengine/system/TypeFileStorage.java
tiramisu24: ^that shows that duplicate storage types will override, which is the problem? We will have duplicate storage types, with different search types (index)

MrCreosote: Unless I'm misunderstanding what you're saying, I don't think it does show that. It shows that duplicate *search types* will override each other for a particular `StorageType`. There's only one `StorageType` per `TypeMapping` instance, but potentially many `SearchTypes`.

lib/src/kbasesearchengine/system/TypeFileStorage.java
MrCreosote: Ok yeah, I misread search type as storage type or got it confused in my head somehow.

> by spec, I meant the yaml files we use to index

The type mapping yaml files or the search type yaml files?

scanon:
I think this looks pretty good.  One minor addition would be a unit test for hasParentId.

tiramisu24:
> I think this looks pretty good. One minor addition would be a unit test for hasParentId.

done

